PR #717, opened 2015-01-17
[WIP] Support for traits - DONT MERGE!!

ovr:
Start working on this


andresgutierrez:
ping @ovr

hylent:
:+1:

baszczewski:
@ovr @andresgutierrez 
Is there any problems with this traits implementation?

steffengy:
It's simply not done by the looks of it, and also does not contain any specification.

maciejmrozinski:
@ovr what is the status of this PR? I would love to see traits landed in Zephir :)

Jurigag:
We need to consider two things that implementing trait for being used for PHP can be sone really east, just pretty much this PR + a few small changes in Library so trait keyword will be recognized and correctly compiled as trait - the biggest issue is that INTERNAL trait itself in php 7 is supported from something like php 7.0.11 and later as i checked - otherwise there is just seg fault, it's php issue and it was fixed on 7.0.11.

Not sure about php 5, it's possible that same thing is happening.

But if we would like to have trait working within zephir itself - there will be much more work to do, because C doesn't have things like traits - we would need to somehow copy methods, headers etc, not sure exactly.

sergeyklay:
👎 for traits. There is more important part to implement and bugs to fix.

Jurigag:
I agree, anyway i think we don't need trait in zephir itself, and implementing them for PHP will be easy job - but as i wrote - the problem is php support for internal trait, the only way i see here is to throw exception traits are not supported on php 7.0.0-7.0.10

maciejmrozinski:
@Jurigag I think that traits implemented in PHP part will be enough for 99% of use cases and most of people using Zephir will be satisfied.

PR #719, opened 2015-01-19
[TDD] Added test for branch and type

ovr:
We need to detect variable type for success branch
It's need to be array type, not variable


andresgutierrez:
ping @ovr

sergeyklay:
Is there any progress?

PR #723, opened 2015-01-22
[NFR] Short Ternary Operator

ovr:


ovr:
ping @steffengy

steffengy:
Merged manually without the regenerated test files since before 
committing all regenerated stuff I want to minimize the differences (which requires restoring old behavior and keeping benefits of the new code, which is a bit tricky)

ovr:
@steffengy 
Nice, thank you :smile_cat:

Issue #725, opened 2015-01-24
Superglobals like _GET not globally mutable


steffengy:
If you do something like below _GET["test"] will only be modified in that specific function,
but when you access it from another scope 
(for example from another function or generally code that runs later), 
the value and therefor the superglobal remains unchanged.

```
public static function test()
{
    let _GET["test"] = 1;
}
```

Could be because of PH_SEPARATE?

ovr:
Please added more tests for all types

steffengy:
@ovr Done. Just didn't get to it yesterday

PR #726, opened 2015-01-24
When assigning a value to a superglobal, check for zval separation

steffengy:
This hopefully addresses the issue demonstrated in #725.

Not sure if that's the correct way to fix this, but it seemed relatively safe.

### DIFF
--- a/Library/Statements/Let/ArrayIndex.php
+++ b/Library/Statements/Let/ArrayIndex.php
@@ -171,6 +171,11 @@ class ArrayIndex
 
         $compilationContext->headersManager->add('kernel/array');
 
+        if ($isGlobalVariable) {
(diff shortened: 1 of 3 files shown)


steffengy:
@ovr 
Anything else I should change on this PR?

ovr:
@steffengy 
I think this can be bugged when you will use

```zep
let _GET[12345] = "my test";
```

steffengy:
@ovr
just like that it seems to work for me:
```
array(1) { [12345]=> string(7) "my test" }
```

Or do you mean a specific case, where it might bug?

ovr:
@steffengy 
I mean this case, seems like okey

ovr:
@phalcon @andresgutierrez Review it please

PR #734, opened 2015-01-28
Fix function cache for static methods

### LINKED ISSUE #721: FunctionCache incorrectly caches static functions calls
After weeks, I finally managed to pinpoint the issue and produce a small testcase:
Starts to appear with e1a78ce707b408d1ce80740890478aef56a764ad
(as i already mentioned, this is the followup)

Zephir Code:

```
namespace Test;

class DebugBase
{
    public function doAction()
    {
        var s;
        let s = "Test\\DebugInstance1";
        echo({s}::returnHello());
        let s = "Test\\DebugInstance2";
        echo({s}::returnHello());
    }
}
```

```
namespace Test;

class DebugInstance1
{
    public static function returnHello()
    {
        return "hello";
    }
}
```

```
namespace Test;

class DebugInstance2
{
    public static function returnHello()
    {
        return "bye";
    }
}
```

PHP-Code:
```
<?php
namespace Test;
$d = new DebugBase();
$d->doAction();
```

Output:
```
hellohello
```

Expected:
```
hellobye
```

(Assumed) Cause of this bug:
https://github.com/phalcon/zephir/blob/3661e14c8495a129841fd11589033956b939647b/ext/kernel/fcall.c#L413

I think the problem is on a static function call the target of the call 
is not used to generate a fcache-hash.

```if (type != zephir_fcall_function && !object_pp) { ``` is automatically true on a static call, 
so object_pp is set and therefor object_ce is ignored and not passed further, 
leading to identicalhash keys and therefor to the 
same cache result. 
Therefore only the calling scope and function name is taken to account.

Hopefully someone who understands the internals, will be able to properly fix this.

steffengy:
Addresses and resolves #721.

A static fucntion call in that case is handled ("transformed" by the compiler) as zephir_fcall_ce.
(ce = class entry, which already suggests that we have no object scope)
The current code still passes object_pp (which is an object scope), insteadof the class entry,
as the call type would suggest.

I'm not sure if there are similar issues, with different call types.

### DIFF
--- a/ext/kernel/fcall.c
+++ b/ext/kernel/fcall.c
@@ -410,7 +410,7 @@ int zephir_call_user_function(zval **object_pp, zend_class_entry *obj_ce, zephir
 
 	if (!cache_entry || !*cache_entry) {
 		if (zephir_globals_ptr->cache_enabled) {
-			fcall_key_hash = zephir_make_fcall_key(&fcall_key, &fcall_key_len, (object_pp ? Z_OBJCE_PP(object_pp) : obj_ce), type, function_name TSRMLS_CC);
+			fcall_key_hash = zephir_make_fcall_key(&fcall_key, &fcall_key_len, (object_pp && type != zephir_fcall_ce ? Z_OBJCE_PP(object_pp) : obj_ce), type, function_name TSRMLS_CC);
 		}
 	}
 


phalcon:
Thanks

PR #735, opened 2015-01-28
Fix the not operator, to respect code expressions

### LINKED ISSUE #680: Incorrect behavior of NOT operator
I have a code:

```zep

	{
		var _array;

		let _array = false;

		var_dump(
			is_array(_array), // false — ok
			!is_array(_array), // false - NOT ok
			!is_array(_array) === true, // false — NOT ok
			is_array(_array) === false, // true - ok
			!is_array(_array) === false, // true - WTF????
			is_array(_array) === !is_array(_array), // false - ok
			is_array(_array) == !is_array(_array) // false - ok
		);
	}
```

It produce the output:

```
bool(false)
bool(false)
bool(false)
bool(true)
bool(true)
bool(false)
bool(false)
```

How it's possible?

Zephir version 0.5.9a [396cb30]

steffengy:
Code like
```
!Z_TYPE_P(
    (_array ? ZEPHIR_GLOBAL(global_true) : ZEPHIR_GLOBAL(global_false))
) == IS_ARRAY
```
could be produced before.

The problem is, that the not operator (!) has a higher precedence than the assignment operator (=),
which leads to a boolean being compared with IS_ARRAY.
This fixes #680.

### DIFF
--- a/Library/Operators/Unary/NotOperator.php
+++ b/Library/Operators/Unary/NotOperator.php
@@ -51,7 +51,7 @@ class NotOperator extends BaseOperator
             case 'uint':
             case 'long':
             case 'ulong':
-                return new CompiledExpression('bool', '!' . $left->getCode(), $expression);
+                return new CompiledExpression('bool', '!(' . $left->getCode() . ')', $expression);
 
             case 'variable':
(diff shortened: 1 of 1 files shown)


ovr:
Seems like okey, thanks

PR #738, opened 2015-01-29
fcache: Only store cache entries, if they are initialized.

### LINKED ISSUE #715: Cache method/function in PHP 5.6
After https://github.com/phalcon/zephir/commit/e1a78ce707b408d1ce80740890478aef56a764ad I'm getting an Segmentation fault, eg. on the [ice/framework](https://github.com/ice/framework) and [ice/website](https://github.com/ice/website) app. 

@ovr after clone the framework please run `composer update`

steffengy:
This should fix #715, 
where somehow some function cache entries are uninitialized and still cached.

This leads to some strange null-pointer dereferences and other similar stuff.
I hope the explanation in the commit itself is detailed enough, so I will leave
that summary here that short.

### DIFF
--- a/ext/kernel/fcall.c
+++ b/ext/kernel/fcall.c
@@ -488,7 +488,7 @@ int zephir_call_user_function(zval **object_pp, zend_class_entry *obj_ce, zephir
 	EG(scope) = old_scope;
 
 	if (!cache_entry || !*cache_entry) {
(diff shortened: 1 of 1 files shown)


ovr:
:+1:

phalcon:
Thanks

Issue #702, opened 2014-12-28
Zephir compile class with incorrect parent class.


serg-deep:
zephir build
Can't register class It\Db\Query\Placeholder::BasePlaceholder with null parent
PHP Fatal error:  Unable to start it module in Unknown on line 0

serg-deep:
OS: debian 6
GCC: 4.4.5 (Debian 4.4.5-8)
Zephir: 0.5.9a

This is arhive of extensgion http://rghost.net/60111229.

Extenshion structure:
```php
| - it
    | - Exception.zep
    | - db
          | - Exception.zep
          | - query
               | - Exception.zep
               | - placeholder
                    | - Exception.zep
```
File it\Exception.zep
```php
namespace It;

use \Exception as PHPException;

class Exception extends PHPException{

}
```

File it\db\Exception.zep
```php
namespace It\Db;

use It\Exception as ItException;

class Exception extends ItException{

}
```

File it\db\query\Exception.zep
```php
namespace It\Db\Query;

use It\Db\Exception as ItDbException;

class Exception extends ItDbException{

}
```

File it\db\query\Placeholder\Exception.zep
```php
namespace It\Db\Query\Placeholder;

use It\Db\Query\Exception as ItDbQueryException; 

class Exception extends ItDbQueryException{

}
```

If u compile it and add to php.ini, u get the error:
```php
Can't register class It\Db::Exception with null parent
PHP Fatal error:  Unable to start it module in Unknown on line 0
```
If delete it\db\query\Placeholder\Exception.zep extension will successful compiled.

serg-deep:
Any activity on this issue?

ovr:
I was trying to fix this, but I cant :anguished: 
Maybe when i was debugging I miss something

@serg-deep
Please load this project on GitHub

ovr:
Closed by https://github.com/phalcon/zephir/pull/751
Thank @steffengy

serg-deep:
Ty )

PR #742, opened 2015-01-30
Make sure that extended/implemented classes are initialized first

steffengy:
This addresses #702.

Currently it's possible that extended/implemented class have the same 
depedency rank, which can lead to the extending class to be initialized
first, where an error occurs becurs the extended class does not exist yet.

I'm not quite sure if that's the best way to do it, so I'm open to feedback.


ovr:
@steffengy 
Can you please add tests? This will be helpful ;)

steffengy:
This version is unfortunately bugged.

Issue #746, opened 2015-02-02
Array assignment of boolean expression always results in false


steffengy:
A class like:
```
public g;
public s = [];

    public function doAction()
    {
        var str = "test";
        let this->s["a"] = str == "abc";
        return this->s;
    }
```

generates:
```
zval *str, *_0;

	ZEPHIR_MM_GROW();
	ZEPHIR_INIT_VAR(str);
	ZVAL_STRING(str, "test", 1);

	ZEPHIR_INIT_VAR(_0);
	ZVAL_STRING(_0, "a", 1);
	zephir_update_property_array(this_ptr, SL("s"), _0, ZEPHIR_GLOBAL(global_false) TSRMLS_CC);
	RETURN_MM_MEMBER(this_ptr, "s");
```

The array property is always set to true (even if the string is dynamic).

ovr:
Closed via your PR :)

PR #747, opened 2015-02-02
Fix assigning a boolean to an array, when a boolean expression is requir...

steffengy:
This addresses #746


ovr:
@steffengy 
Can you please add a small test case?

steffengy:
@ovr sure, is that sufficient?

ovr:
Thank you

PR #752, opened 2015-02-11
Update Compiler.php

ghost:
修复windows系统无法直接执行JS的问题


steffengy:
I dislike this for various reasons:
If you run it multiple times your file will contain the same line for n times. 
(Where n is the amount of times run, after it was initially created)

It also does not make sense to create the configure.bat, 
because that is a file that is created by phpize.

What is the reasoning behind these changes?

xutongle:
@steffengy 我在Windows 8.1 VC11下编译提示我无法执行JS，参考PHP的编译脚本发现用BAT来启动JS是可以使用的。

PR #759, opened 2015-02-14
Fix DivOperator to be thread-safe by adding missing TSRMLS_CC

steffengy:
Trivial fix, to allow compiling as ZTS, when using the div operator.


ovr:
Thank you

Issue #517, opened 2014-09-09
Functions in namespace


wapmorgan:
It does not support standalone functions in namespaces.
Please, consider providing this feature.
**Zephir, test/functions.php**
```php
namespace test;

function formatOutput(text) {
    // .. some stuff
}
```

**PHP 5.3**
```php
echo test\formatOutput('example');
```

steffengy:
@ovr 
this is implemented now, issue should be closed.

serg-deep:
Thank you. This is really great feature.

PR #751, opened 2015-02-08
Make sure that extended/implemented classes are initialized first

steffengy:
This should ensure that dependencies are loaded in the correct order
by recursively setting the rank accordingly

Followup of #742 addressing #702

This seems to work so far, needs some testing

PR #762, opened 2015-02-16
Implement global and namespace(d) function declaration

steffengy:
Addresses #517 
This is not intended to allow bypassing the OOP pattern, but to aid creating helper functions:
----

For example if you use PHP Code and want to define an 
language *helper* ```l($str)```
to return the localized version of a string.

This is  - for example - incredibly useful for the usage in templates.

- Sure you could also use a normal Method like ```App::l($str)```
which is really annoying to type every time.

- or workaround by using PHP to define something like
```function l($str) { return App::l($str); } ```

But a native implementation as addition to zephir's
useability is much **more comfortable**, because you can just declare the helper within
zephir.


ovr:
Awesome :+1:

steffengy:
@ovr 
what do you mean?
If you declare a function before the namespace declaration,
it will be declared as **global**.
If you declare it after the namespace declaration,
it will be declared **in said namespace**

I think the test cases I Added explain it pretty good: 
https://github.com/steffengy/zephir/blob/08ccaf120d2cdc0d5953b9d8788127615f0fbadc/test/fcall.zep#l7-17
https://github.com/steffengy/zephir/blob/08ccaf120d2cdc0d5953b9d8788127615f0fbadc/unit-tests/Extension/FcallTest.php#l56-57

ovr:
@steffengy 
Thank for answer

steffengy:
@ovr
No problem :)
Unfortunately, It will hopefully not take very much time to get this reviewed and (hopefully) merged.

ovr:
@steffengy 

kk, I will review it ASAP

Will merge this with 
#765 
#751 
and release `0.6`

steffengy:
@ovr 
Sounds good :)

ovr:
Thank you

mruz:
@steffengy can I add checking if global function with this name exist before?

steffengy:
@mruz 
Yeah I think currently it only checks that per-file.
I think for a namespace/global scope it's still missing (should be somewhere in Compiler)

PR #766, opened 2015-02-17
Array Assign: Fix reference copy to protect the source from the GC

### LINKED ISSUE #709: Incorrect array element fetching on var assignment
I have a simple code:

```zep

	{
		var arr;

		let arr = [1, 2];
		let arr = arr[array_rand(arr)];

		echo arr, "\n";
	}

```
But it produce incorrect output. I ran it several times and got this:
```
4478337920
4513325952
2
4422243200
4441392000
4582888320
2
2
4422804352
4594230144
2
2
4509430656
...
```

If I create an additional variable like that:

```zep
	{
		var arr, arrElem;

		let arr = [1, 2];
		let arrElem = arr[array_rand(arr)];

		echo arrElem, "\n";
	}
```

all works fine.

Zephir version 0.5.9a

steffengy:
Fixes #709.

The problem is that it's possible that the garbage collector frees the data which should be copied.

Problem analysis:
```
zephir_array_fetch(&_1, arr, _2, PH_NOISY | PH_READONLY, "xxx", 59 TSRMLS_CC);
ZEPHIR_CPY_WRT(arr, _1);
```
The problem is that _1 is potentially a pointer to an element of arr 
and ZEPHIR_CPY_WRT does add the reference AFTER calling
the Destructor (which may lead to garbage collector activity
and therefore to an "use-after-free" bug.)

### DIFF
--- a/ext/kernel/memory.h
+++ b/ext/kernel/memory.h
@@ -160,6 +160,7 @@ void zephir_deinitialize_memory(TSRMLS_D);
 	}
 
 #define ZEPHIR_CPY_WRT(d, v) \
+	Z_ADDREF_P(v); \
 	if (d) { \
 		if (Z_REFCOUNT_P(d) > 0) { \
 			zephir_ptr_dtor(&d); \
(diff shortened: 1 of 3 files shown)


ovr:
Seems to be okey
Thank you :smiley:

PR #770, opened 2015-02-21
Travis CI - compile under clang

ovr:
Clang is better than gcc
We need to support it on CI service


andresgutierrez:
ping @ovr

sergeyklay:
Implemented in the `master` branch

PR #804, opened 2015-03-04
Allow build on PHP 7

mruz:
When I'm trying to build with travis on PHP 7:
> phalcon/zephir dev-master requires php ~5.3 -> your PHP version does not satisfy that requirement.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "issues": "https://github.com/phalcon/zephir/issues?state=open"
     },
     "require": {
-        "php": "~5.3",
+        "php": ">=5.3",
         "ext-json": "*",
         "ext-hash": "*",
         "ext-ctype": "*"


steffengy:
Why would you try to use zephir with PHP7?
It just does not compile, so why would you want to run tests?
Or do you just use the zephir compiler/composer with PHP7?

(And no upgrading zephir to PHPNG is not very trivial, see the issue on the matter)

mruz:
I'm aware of this, I just wanted to see what stops the compilation.

steffengy:
Fundamental changes to the zend engine ^^
(They require fundamental changes (weeks of work atleast)
to zephir, I will probably release my experiment soonish,
which will provide atleast 100% working tests of zephir)

Enabling tests for something that only fails (and cannot work) does just not make sense.
That's something that should be done when implementing PHP7 support.

hylent:
:+1:

golovanov:
:+1:

ovr:
@steffengy You are welcome (help with PHP 7 support)

ovr:
Thank you @mruz

PR #806, opened 2015-03-07
Fixed switch behaviour, fixed #801

### LINKED ISSUE #801: Incorrect Switch behavior
I have a simple code:

```zep
	{
		var variant;

		let variant = 1;

		switch (variant) {
			case 1:
			default:
				return 1;

			case 2:
				return 2;

			case 3:
				return 3;

			case 4:
				return 4;
		}
	}
```

This function by some reason return 2 instead of 1. I can't explain it, but function below works fine:

```zep

	{
		var variant;

		let variant = 1;

		switch (variant) {
			case 2:
				return 2;

			case 3:
				return 3;

			case 4:
				return 4;

			case 1:
			default:
				return 1;
		}
	}

```

Zephir 0.6.0a

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9115571-incorrect-switch-behavior?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nkt:


ovr:
Seems to be okey
Thank you

Issue #812, opened 2015-03-12
Input method argument with type `uint` not coverted to uint.


kse300489:
Example:

``` php
public function simpleMethod(const uint! val){
  var_dump(val);
}
```

Run this code:

``` php
$this->simpleMethod(1);
$this->simpleMethod(-1);
```

Output:

``` text
int(1)
int(-1)
```

Must be:

``` text
int(1)
int(1)
```

Or throw error, that input argument not unsigned int.
##

PR #819, opened 2015-03-15
Added tests for #812

nkt:


unit-tests/Extension/MCallTest.php
```diff
@@ -173,10 +164,7 @@ public function arrayParam()
         $this->assertTrue($t->arrayParam(array(1, 2, 3)) === array(1, 2, 3));
     }
 
-    /**
-     * @test
```
ovr: It's needed to work with hack method
`getMethodFirstParameter`

andresgutierrez:
ping @nkt

Issue #824, opened 2015-03-17
Syntax error when trying to build Zephir


BonnieDoug:
Just gone through the install steps for Zephir, but once I get to the Zephir Build step I get the following error:

    c:\zephir>zephir build
    Zephir\ParseException: Syntax error in C:\zephir\test\fcall.zep on line 7

        function zephir_global_method_test(var str)
        -------^

    at Library\CompilerFile.php(534)
    #0 Library\Compiler.php(159): Zephir\CompilerFile->preCompile(Object(Zephir\Compiler))
    #1 Library\Compiler.php(199): Zephir\Compiler->preCompile('test\fcall.zep')
    #2 Library\Compiler.php(686): Zephir\Compiler->recursivePreCompile('test')
    #3 Library\Compiler.php(880): Zephir\Compiler->generate(Object(Zephir\Commands\CommandBuild))
    #4 Library\Compiler.php(985): Zephir\Compiler->compile(Object(Zephir\Commands\CommandBuild), false)
    #5 Library\Compiler.php(1054): Zephir\Compiler->install(Object(Zephir\Commands\CommandBuild), false)
    #6 Library\Commands\CommandAbstract.php(72): Zephir\Compiler>build(Object(Zephir\Commands\CommandBuild))
    #7 Library\Bootstrap.php(200): Zephir\Commands\CommandAbstract->execute(Object(Zephir\Config), Object(Zephir\Logger))
    #8 compiler.php(21): Zephir\Bootstrap::boot()
    #9 {main}

This is using php 5.6 nts with VS2012.

System PATH currently has these (Plus many more)

    C:\xampp\php; //(5.5 nts)
    C:\Program Files\Microsoft SQL Server\110\Tools\Binn\;
    C:\zephir\bin;
    C:\phpdevpack; //(php-5.6.6-nts-Win32-VC11-x86.zip)
    C:\phpsdk; //(php-sdk-binary-tools-20110915.zip)

Not to sure what else to put here but I'll happily supply any other info needed.

ovr:
Did you recompile a parser?

BonnieDoug:
All I've done is follow the steps as described here https://github.com/phalcon/zephir/blob/master/WINDOWS.md except for the setx commands were failing with a "WARNING: The data being saved is truncated to 1024 characters." error, so I set them manually in System/Advanced System Settings/Environmental Variables/Path

andresgutierrez:
You need a visibility before 'function': public, protected, private, etc

[...]

PR #826, opened 2015-03-19
Rebuild the Win32 Parser (VC2012) and remove unnecessary cygwin binaries

steffengy:
Also update json-c to newest revision and add build instructions.

Should address #824


steffengy:
@ping ovr

andresgutierrez:
Thanks

BonnieDoug:
+1 will give it a go tonight :-)

steffengy:
Yeah but the WINDOWS.md assumes that you use the already built parser.
(because building the parser is annoying)
The section about building the parser is just for information,
if you want to do it yourself.

BonnieDoug:
Ah, I get it. Still learning the ropes here, think I've been a bit spoilt in the past with symfony holding your hand every step of the way (at the expense of speed). Just bought "head first : C" so hopefully I'll get a bit more adept and actually contribute rather than leeching :-)

steffengy:
Yeah I admit that it's not very intuitive,
but I just added the instructions there quickly, 
that if somebody wanted to build it, he could, even though he wouldn't need to.

It should be definitely marked as optional with explanation,
that you could also just use the .exe from the repository :)

PR #831, opened 2015-03-20
Implement/Enable some optimizers (string, hashing and time related)

steffengy:
Implemented some optimizers
- ceil
- floor
- crc32
- microtime
- time
- pow

Added some optimizers for already existing wrappers:
- md5
- ucfirst


ovr:
Good job, thank you

PR #832, opened 2015-03-21
Fix ExitDieTest to use the PHP binary of zephir (which runs phpunit)

steffengy:
This then also allows all tests to pass on Win32,
when using a non-debug build.

This also generally allows to have another PHP binary in path
than the one using zephir.


ovr:
thank you

Issue #833, opened 2015-03-21
SubstrOptimizer disabled, what bugs has the implementation?


steffengy:
The substring optimizer is disabled since 1ba122246415bc4f79eabe4012bef975e4050eb8
which isn't quite optimal, since it's one of the most common functions.

A comment says it's implementation has bugs:

Does anybody know **what** bugs?

@andresgutierrez 
Do you recall what bugs?

andresgutierrez:
It does not work passing an array as first parameter

steffengy:
@andresgutierrez 
PHP's behavior is to just throw a warning and return null:
```substr() expects parameter 1 to be string, array given on line```
current behavior (with pached #834) is to cause:
```Notice: Array to string conversion```

should that be adjusted?

PR #834, opened 2015-03-21
Fix ZTS for time and more work on optimizers (Substr, Round)

steffengy:
Fixed the ZTS issue with the **microtime optimizer**, which caused the travis failure.
(Unfortunately I didn't think of that issue, so it sneaked into my PR...)

Fixed the **substr optimizer** and added some of PHP's own unit tests for it.
Implemented a round optimizer.

Added an optimizer for **is_int** and made 
**zephir_fetch_nproperty_this_quick** use the passed hash
instead of calculating one again.

### DIFF
--- a/ext/kernel/operators.c
+++ b/ext/kernel/operators.c
@@ -24,6 +24,7 @@
 
 #include "php.h"
 #include "ext/standard/php_string.h"
+#include "ext/standard/php_math.h"
 #include "php_ext.h"
 #include "kernel/main.h"
 #include "kernel/memory.h"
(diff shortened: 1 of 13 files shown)


steffengy:
addresses #833 and hopefully fixed all of the "mentioned" bugs.

ext/kernel/operators.c
```diff
@@ -893,6 +894,42 @@ void zephir_ceil(zval *return_value, zval *op1 TSRMLS_DC)
 	RETURN_FALSE;
 }
 
+extern double _php_math_round(double value, int places, int mode);
+void zephir_round(zval *return_value, zval *op1, zval *op2, zval *op3 TSRMLS_DC)
+{
+    int places = 0;
+	long mode = PHP_ROUND_HALF_UP;
```
ovr: Can you fix it?

ext/kernel/operators.c
```diff
@@ -893,6 +894,42 @@ void zephir_ceil(zval *return_value, zval *op1 TSRMLS_DC)
...
+    convert_scalar_to_number_ex(&op1);
+
+    if (op2) {
+        places = zephir_get_intval_ex(op2);
+    }
+    if (op3) {
+        mode = zephir_get_intval_ex(op3);
+    }
+
+	switch (Z_TYPE_PP(&op1)) {
```
ovr: Can you fix? I mean spaces
We are using tabs
https://github.com/phalcon/zephir/blob/master/.editorconfig#L15

steffengy:
@ovr 
Seems like my editor was a bit rebellious and just ignored the editorconfig,
sometimes happens, sorry for that.

I also added some trivial changes 
(is_int optimizer and zephir_fetch_nproperty_this_quick fix)

If anything else has to be changed, please let me know :)

andresgutierrez:
Thanks

PR #836, opened 2015-03-22
FCache: use emalloc instead of ecalloc for the fcall-key buffer

steffengy:
The buffer is anyways completely overwritten by the 3 memcpys,
that's also how the needed length is calculated.
So there's no need to zero-out the allocated memory

I guess that's okay, if there wasn't any other reason for using ecalloc.


ovr:
cC @andresgutierrez

ovr:
ping @andresgutierrez

steffengy:
@andresgutierrez 
Can you take a look at this and elaborate on the choice of using ecalloc?

PR #850, opened 2015-03-29
Added php 7.0 and hhvm

vieirasantosn:
Added php7 and hhvm to build and allow them to fail


steffengy:
I would really like to understand why people want to allow builds for things that can't work...
-1

PR #857, opened 2015-03-31
Travis Fix: PSR2-Fix to allow travis build to pass again

steffengy:
Fix travis to compile again (after fcall optimizations)


andresgutierrez:
Thanks

PR #858, opened 2015-03-31
Fix travis build: re2c exception

steffengy:
The re2c compilation step fails since c++ comments ``//`` are used.

Replaced them with C-style `/**/` block-comments and now
the parsing seems to work, resulting in a successful travis build.

https://travis-ci.org/steffengy/zephir/builds/56620190


valVk:
Thanks

PR #866, opened 2015-04-03
Multi-Array Update: Remove code duplication for objects and simple array

steffengy:
Before there was pretty much the same code  at 3 places.
Now they *share* an implementation, increasing maintainability.

The multi-array implementation itself unfortunately is not
cleaned up yet, but is surely in need of large amounts
of refactoring. 

(IMO it's really ugly to read and understand since there is much code duplication)


steffengy:
@andresgutierrez 
Could you review this please?

andresgutierrez:
Thanks

PR #869, opened 2015-04-04
Improvement: Make property access on non-objects easier to find 

steffengy:
Adjust the notice messages to include the property name to make errors
easier to locate.


andresgutierrez:
Thanks

Issue #772, opened 2015-02-23
Potential Duplicate Code in Zephir\Compiler::generate() 


pjacferreira:
Hi, 

As per a request by zyxep (from IRC) I have created this issue.

I was doing a little debugging of the PHALCON 2 Zephir Code and noticed something strange.
In the method Zephir\Compiler::generate() I noticed that lines 686 - 714 are repeated (again) at positions 749 - 778.

I wasn't sure if this was intentional (i.e. a 2 pass compiler system) or not?

ovr:
Thank for feedback
Will be reviewed ASAP

ovr:
Fixed by @steffengy

PR #870, opened 2015-04-04
Add checks to detect whether a function has been declared multiple times

steffengy:
Previously it lead to compile-time errors, when a function was (invalidly) declared multiple times.

This still supports multiple function declarations (with the same name) in multiple namespaces.
It detects and fails on:
- multiple global functions with the same name
- multiple namespaced functions, with the same name in the same namespace

The code duplication mentioned in #772 is also removed.

### DIFF
--- a/Library/Compiler.php
+++ b/Library/Compiler.php
@@ -139,6 +139,22 @@ class Compiler
             }
         }
     }
+    
+    /**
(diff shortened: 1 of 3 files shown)


andresgutierrez:
Thanks

PR #871, opened 2015-04-04
Fix some tests to be actually run

steffengy:
Some tests were simply not run, 
because they did not include the "Test" affix or were wrongly named.

The ArrayObject Test also expected invalid behavior,
which wasn't equal to PHP's behavior.


ovr:
Thank you, appreciate you work ;)

PR #872, opened 2015-04-05
Removed a leading underscore from a protected property.

SidRoberts:
This was left out of https://github.com/phalcon/zephir/commit/247072464e571b0d6b1ec19bc20c362859fbd125#diff-1


ovr:
Thank you

PR #875, opened 2015-04-06
Fix win32 build: wrong position of ZEPHIR_FASTCALL at zephir_create_arra...

steffengy:
Trivial change, just a ZEPHIR_FASTCALL at the wrong position, which confuses MSVC


ovr:
Thank you

PR #877, opened 2015-04-07
Better method cache analysis

andresgutierrez:
- Analyses how many times a method is called in a specific scope and adds an inline cache
- Checks if the caller belongs to a class of the same extension or internal class and adds an inline cache
- Checks if the method is called within a loop without any mutation and adds an inline cache
- Checks if the method is final or private and adds an inline cache
- Checks if the class is final and adds an inline cache


ovr:
Awesome, good job :+1:

PR #890, opened 2015-04-15
Add some missing json and pcre ext checks for Win32 builds

steffengy:
Make sure json_encode/decode calls
(and maybe later preg_match calls, whatever the issues with the optimizer is)
are also really optimized on win32.


andresgutierrez:
Thanks

PR #891, opened 2015-04-15
Preg-match optimizer: Re-enable with fixes

steffengy:
- Also added some additional tests
- Fixed the fallback implementation, when no PCRE is available during compilation.

Any other issues the current implementation has, which have to be fixed?


steffengy:
@andresgutierrez 
Anything else wrong with the preg_match optimizer?

PR #895, opened 2015-04-19
Fix #894: Static property access may observe non-tracked temp variables

### LINKED ISSUE #894: Static property access causes read-only (not tracked) temp variables to be observed
I experience property fetching randomly failing (this_ptr corrupted or something?)
If that's the case zephir_read_property or fetch_property_this results in an invalid zval,
that has nothing to do with the actual value of the actual property. 
(IS_ARRAY insteadof IS_LONG - eventually rarely something completely uninitialized)

This is all resolved by commeting the new 50600 function call optimization ifdefs:
State: Investigating...

* Bug caused by: zephir_fcall_method/ZEPHIR_FCALL_TYPE_ZVAL_METHOD *
* Disappears when doing the following in **zephir_call_class_method_aparams**:
```
case zephir_fcall_method:
    default:
        assert(object != NULL);
      + Z_ADDREF_P(object);
``` 
(but results in a memleak in another case)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11294555-static-property-access-causes-read-only-not-tracked-temp-variables-to-be-observed?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github).
</bountysource-plugin>

steffengy:
This bug was quite annoying to track down,
since unfortunately it corrupted values far before it crashed.

Details are in the respective issue, finally getting some sleep after hours of investigating.

### DIFF
--- a/Library/Expression/StaticPropertyAccess.php
+++ b/Library/Expression/StaticPropertyAccess.php
@@ -154,15 +154,8 @@ class StaticPropertyAccess
 
         $compilationContext->headersManager->add('kernel/object');
 
-        if ($classDefinition == $compilationContext->classDefinition) {
-            if ($this->_readOnly) {
-                $compilationContext->codePrinter->output($symbolVariable->getName() . ' = zephir_fetch_static_property_ce(' . $classDefinition->getClassEntry() . ', SL("' . $property . '") TSRMLS_CC);');
-            } else {
-                if ($symbolVariable->getName() != 'return_value') {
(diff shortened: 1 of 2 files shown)


andresgutierrez:
Thanks!

PR #902, opened 2015-04-20
Fix zephir_floor and zephir_ceil to be consistent and not buggy

steffengy:
Previously ``convert_scalar_to_number_ex`` was able to cause zval corruption,
when a string was passed as operand. The implementation now uses zephir
internal macro's and works as intended (without affecting the passed operand).


andresgutierrez:
Thanks

PR #904, opened 2015-04-20
[NFR] Added prototypes disabled option

kjdev:
Because there could be an unexpected error such as #900,
'prototypes' or not be better there was an option to disable.

```
zephir generate --prototypes=disabled  # prototypes unloaded
```

### DIFF
--- a/Library/Compiler.php
+++ b/Library/Compiler.php
@@ -768,7 +768,17 @@ class Compiler
                 }
             }
 
-            if (is_dir(ZEPHIRPATH . 'prototypes') && is_readable(ZEPHIRPATH . 'prototypes')) {
+            $prototypes = $this->config->get('prototypes', 'extra');
+            if ($prototypes !== null
+                && (empty($prototypes)
+                    || strcasecmp($prototypes, 'no') == 0
+                    || strncasecmp($prototypes, 'disable', 7) == 0)) {
+                $prototypes = false;
+            } else {
+                $prototypes = true;
+            }
+
+            if ($prototypes == true && is_dir(ZEPHIRPATH . 'prototypes') && is_readable(ZEPHIRPATH . 'prototypes')) {
                 /**
                  * Load additional extension prototypes
                  * @var $file \DirectoryIterator


andresgutierrez:
I think to make it work you have to add the arguments it receives

kjdev:
If you specify a `disable` or` no` the `--prototypes` in zephir options, `ZEPHIRPATH  . 'protetypes` have to avoid loading.

andresgutierrez:
What do you want to fix by adding this option?

kjdev:
Don't want to load the Class that are defined in the protetypes. (Library/Compiler.php:L.780)
It's an error if you attempt to use other than those defined here.

```
771: if (is_dir(ZEPHIRPATH . 'prototypes') && is_readable(ZEPHIRPATH . 'prototypes')) {
772:     /**
773:      * Load additional extension prototypes
774:      * @var $file \DirectoryIterator
775:      */
776:     foreach (new \DirectoryIterator(ZEPHIRPATH . 'prototypes') as $file) {
777:         if (!$file->isDir()) {
778:             $extension = str_replace('.php', '', $file);
779:             if (!extension_loaded($extension)) {
780:                 require $file->getRealPath();
781:             }
782:         }
783:     }
784: }
```

hylent:
If using some method which is in some extension but not in the zephir prototype class, there will be errors

andresgutierrez:
@kjdev why not update the prototypes instead to match the signature in the extension?

ovr:
ping @kjdev

kjdev:
* prototypes implementation of incomplete (not exhaustive)
* to match the expansion of the update must be maintained

We can not build in the incomplete implementation when the zephir you tried to use other than phalcon is troubled.

andresgutierrez:
@kjdev if you have the extension loaded it will ignore the prototypes

kjdev:
@andresgutierrez It has confirmed that to ignore the prototype if the extension is loaded.
Build in some way at this time is we would like to avoid to stop even if in the build environment is not installed the extension.

KorsaR-ZN:
:+1:

PR #910, opened 2015-04-23
Fixed #740 issue

### LINKED ISSUE #740: Segmentation fault when used throw + sprintf [$10]
How to reproduce:

```zephir
namespace Foo;

class Bar
{

    public function test1()
    {
        echo __CLASS__;
    }

    public function test2()
    {
        throw sprintf("%s\n", __CLASS__);
    }
}

```

Tests:

```ssh
$ php -r '$utf = new Foo\Bar; $utf->test1() . PHP_EOL;'
$ Bar
```

Hmm.. Strangely, I was expecting `Foo\Bar`

```ssh
$ php -r '$utf = new Foo\Bar; $utf->test2() . PHP_EOL;'
```

Segmentation fault (core dumped)

<bountysource-plugin>
---
There is a **[$10 open bounty](https://www.bountysource.com/issues/8142666-segmentation-fault-when-used-throw-sprintf?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nkt:
Inside compiler `sprintf` function returns variable with `variable` type.
This patch catch `sprintf` calls and convert them into `new Exception(sprintf(...))`.

### DIFF
--- a/Library/Statements/ThrowStatement.php
+++ b/Library/Statements/ThrowStatement.php
@@ -84,12 +84,12 @@ class ThrowStatement extends StatementAbstract
         $throwExpr = new Expression($expr);
         $resolvedExpr = $throwExpr->compile($compilationContext);
 
-        if ($resolvedExpr->getType() != 'variable') {
+        if (!in_array($resolvedExpr->getType(), array('variable', 'string'))) {
             throw new CompilerException("Expression '" . $resolvedExpr->getType() . '" cannot be used as exception', $expr);
         }
 
         $variableVariable = $compilationContext->symbolTable->getVariableForRead($resolvedExpr->getCode(), $compilationContext, $expr);
-        if ($variableVariable->getType() != 'variable') {
+        if (!in_array($variableVariable->getType(), array('variable', 'string'))) {
             throw new CompilerException("Variable '" . $variableVariable->getType() . "' cannot be used as exception", $expr);
         }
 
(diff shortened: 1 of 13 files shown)


andresgutierrez:
What is the backtrace of this segfault?

nkt:
@andresgutierrez I didn't get any segfault. Actually it's weird, I've tested this case before PR and got:
```
Uncaught exception 'RuntimeException' with message 'Trying to call method getline on a non-object' in php shell code:1
```
Instead of segfault.

andresgutierrez:
I think the solution is not clean, maybe would be better to check here:

https://github.com/phalcon/zephir/blob/master/ext/kernel/exception.c#L55

if variable object is not an object and then create an exception there:

```c
if (Z_TYPE_P(object) != IS_OBJECT) {
    object_copy = object;
    ALLOC_INIT_ZVAL(object);
    object_init_ex(object, zend_exception_get_default(TSRMLS_C));
    // ... call the constructor passing object_copy as parameter
}
```

This will convert any non-object to an exception

nkt:
@andresgutierrez yep, you're right, I'll fix it.

nkt:
@andresgutierrez yep, it works fine. Should I rebase this?

andresgutierrez:
Yes please

nkt:
@andresgutierrez done

nkt:
@andresgutierrez ping

PR #919, opened 2015-04-25
Fix fcall info keys to not collide with functions for static calls

steffengy:
For pre-php5.6 the type of these calls was already included in the function name:
```
case zephir_fcall_parent: add_next_index_stringl(fn, ZEND_STRL("parent"), 1); break;
case zephir_fcall_self:   assert(!ce); add_next_index_stringl(fn, ZEND_STRL("self"), 1); break;
case zephir_fcall_static: assert(!ce); add_next_index_stringl(fn, ZEND_STRL("static"), 1); break;
```

With php5.6 it might happen, that something like
``self::date()`` calls the normal function ``date()``,
if some scope's are setup unluckily and ``date()`` was called previously
and therefore registered the colliding cash entry.

### DIFF
--- a/ext/kernel/fcall.c
+++ b/ext/kernel/fcall.c
@@ -302,12 +302,23 @@ static ulong zephir_make_fcall_info_key(char **result, size_t *length, const zen
 			buf[len - 1] = '\0';
 			break;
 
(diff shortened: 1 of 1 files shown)


andresgutierrez:
Thanks

PR #929, opened 2015-05-02
Fix superglobal-access not to cause warnings, since they are initialized

steffengy:
Fix these annoying branch warnings that a variable was initialized in a
conditional branch, which occured when using super globals
in if-statements.

### DIFF
--- a/Library/SymbolTable.php
+++ b/Library/SymbolTable.php
@@ -193,11 +193,11 @@ class SymbolTable
                  */
                 $compilationContext->codePrinter->output('zephir_get_global(&' . $name . ', SS("' . $name . '") TSRMLS_CC);');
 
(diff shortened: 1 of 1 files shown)


ovr:
Can you add a test case for it? (use super global) to see warning
Thanks

steffengy:
You just need to use something really basic like:
```
if unlikely isset _GET["id"] {
    let _POST["id"] = _GET["id"];
}
if isset _POST["id"] {
    echo "g";
}
```

steffengy:
@andresgutierrez 
Does "not tested" in that context mean that the solution hasn't been tested yet,
or that tests are missing?

If tests are missing, how do you test this specific issue, because
you would basically have to ensure that specific warnings do not appear.

PR #930, opened 2015-05-02
Revert f58a4db5cac21d2b77e8af407fe42a0a9cde0c2b partially

steffengy:
A gatherer is not required, it can be disabled or not needed
(no statements), which causes an exception -> zephir crash

(It being null in these cases is therefor perfectly valid)


ovr:
Thank you, but I am going to suggest
Add `= null` by default to allow pass `null` values when We disable this optimizer

steffengy:
Seems to work like this, PR adapted.

ovr:
Thank you

PR #943, opened 2015-05-07
CastOperator implement (int) cast from string and var string

ovr:


ovr:
ping @andresgutierrez

Green-Cat:
:+1:

PR #956, opened 2015-05-16
Global static cache

andresgutierrez:
Previous static-variable based cache in Zephir had problems in PHP-FPM or when functions/methods were overriden and it also wasn't available on Windows. This new cache implements a slot-based cache where functions and methods candidates of being cached are stored in a global indexed cache which should be safer than the static-variable based cache.
- Methods marked as final/private are candidates to use this cache
- Functions that are always bundled with PHP also benefit from this


ovr:
Before
```
self::f()          2.245
self::f() -> true  2.285
```

Now

```sh
self::f()          2.165
self::f() -> true  2.218
```

PR #958, opened 2015-05-18
Low overhead fcall (only PHP 5.6)

andresgutierrez:
This PR introduces a specialized function to call functions in the PHP userland removing many internal intermediate structures. `zephir_call_func_aparams_fast` is only used when an inline local cache is available for the method. This is when a function is called more than once in the same method or a function is called within a loop.

The following code:

``` php
public static function guid()
{
        return sprintf("%04X%04X-%04X-%04X-%04X-%04X%04X%04X", 
            rand(0, 65535), rand(0, 65535), 
            rand(0, 65535), rand(16384, 20479), 
            rand(32768, 49151), rand(0, 65535), 
            rand(0, 65535), rand(0, 65535));
}
```

Benchmark (executed 1'000.000 times):

``` php
v0.6.2 3289 milliseconds
v0.6.3 2149 milliseconds
65% improvement
```

### DIFF
--- a/ext/kernel/extended/fcall.c
+++ b/ext/kernel/extended/fcall.c
@@ -32,99 +32,6 @@
 #include "kernel/exception.h"
 #include "kernel/backtrace.h"
 
-/**
- * Calls a function/method in the PHP userland
- */
-int zephir_call_user_function_fast(zval **retval_ptr_ptr, zephir_fcall_cache_entry **cache_entry, zend_uint param_count, zval *params[] TSRMLS_DC)
-{
-	zval ***params_ptr, ***params_array = NULL;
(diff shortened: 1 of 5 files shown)


sergeyklay:
:+1:

ovr:
Nice work :+1:

Green-Cat:
:+1:

sergeyklay:
Benchmark for php/hhvm http://3v4l.org/luMgY

pasmanp:
this only apply to php 5.6.x ?

nkt:
Great! :+1:

nkt:
I guess we should update version.

PR #962, opened 2015-05-25
Introducing internal methods

andresgutierrez:
This PR introduces static dispatch in which the implementation of a polymorphic operation is selected at compile-time. A new visibility modifer called `internal` is introduced. Methods marked as `internal` are only visible to the extension, this allows the compiler to generate direct C-calls to these methods instead of using the dynamic dispatch method in the PHP userland that sometimes is slow.

Internal methods are not visible by PHP, this means that a PHP script cannot call these methods. They aren't visible by the PHP reflection system and backtraces do not show these methods as part of the stack, also, some functions like `func_get_args()` and `func_num_args()` do not work in these kind of methods.

In the other hand, internal methods are really fast, for instance doing a recursive fibonacci vs plain php:

``` zep
class McallInternal
{
    internal function fibonacci(long n) -> long
    {
        if n == 1 || n == 2 {
            return 1;
        }
        return this->fibonacci(n - 1) + this->fibonacci(n - 2);
    }

    public function callFibonacci(long n) -> long
    {
        return this->fibonacci(n);
    }
}
```

``` php
<?php

class Math
{
    function fibonacci($n)
    {
        if ($n == 1 || $n == 2) {
                return 1;
        }
        return $this->fibonacci($n - 1) + $this->fibonacci($n - 2);
    }
}

$t = microtime(true);
(new Test\Mcallinternal)->callFibonacci(31);
echo 'Zephir: ', microtime(true) - $t, PHP_EOL;

$t = microtime(true);
(new Math)->fibonacci(31);
echo 'PHP 5.6: ', microtime(true) - $t, PHP_EOL;
```

Results:

``` bash
Zephir: 0.12126803398132
PHP 5.6: 0.613104820251464

Zephir: 0.12285614013672
PHP 5.6: 0.55841708183289

4.5/5x 450%/500% improvement
```

I think this can also be implemented later in properties


sergeyklay:
:+1:

hylent:
:+1:

ovr:
:+1:

Green-Cat:
:+1:

Jurigag:
Could it be possible to introduce internal classes? @andresgutierrez

PR #963, opened 2015-05-25
Do not define internal methods as PHP methods in generated headers

steffengy:
Without the patch this leads to not resolved external symbols.


andresgutierrez:
Thanks

PR #964, opened 2015-05-25
Internal functions: Implement static call, fix unused return value calls

steffengy:
Implement internal methods for static context like ``self::internalFunc`` (only self/static)
Also fix a null pointer dereference when the return value is not used further.


andresgutierrez:
Thanks

Issue #367, opened 2014-05-02
Cannot define static property with default value: array [$10]


ariskemper:
Getting following error when compiling:
Cannot define static property with default value: array

Zephir version: 0.4.1a

Example:

```zep
class Foo
{
    const OP_TYPE_CREATE = "create";
    const OP_TYPE_INDEX  = "index";
    const OP_TYPE_DELETE = "delete";
    const OP_TYPE_UPDATE = "update";

    /**
     * @var array
     */
    public static opTypes = [
        self::OP_TYPE_CREATE,
        self::OP_TYPE_INDEX,
        self::OP_TYPE_DELETE,
        self::OP_TYPE_UPDATE
    ];

    /**
     * @param string opType
     * @return bool
     */
    public static function isValidOpType(string opType) -> boolean
    {
        return in_array(opType, self::opTypes);
    }
}
```

steffengy:
@ovr @andresgutierrez 
This and the other issue now indeed should be closed :)
(dont quite get why the merged commit didn't...)

Issue #188, opened 2014-02-14
[NFR] Add support static property default empty array


pigochu:
class A {
	private static _B = [];

Zephir\CompilerException: Unknown default type: empty-array

ovr:
problem that php core can't do it https://github.com/php/php-src/blob/PHP-5.5.9/Zend/zend_API.c#L3455

fezfez:
Close ?

PR #970, opened 2015-05-28
Implement arrays as default values of static properties

steffengy:
This implements #367 and #188.

This is similarily implemented as for normal properties,
which simply use the "__construct" method to achieve
this property initialization.

For static properties instead we create a new initialization function,
which is executed on initialization (RINIT, Request Init), which
then simply initializes the properties.


Green-Cat:
:+1:

andresgutierrez:
Nice!

PR #971, opened 2015-05-28
Fix __CLASS__ and __NAMESPACE__ to be properly escaped

steffengy:
At the moment if you do something like ``var_dump(__CLASS__)``
you get something like ``DebugNS`` insteadof ``Debug\NS``


ovr:
I think we need `tests` on it

PR #972, opened 2015-05-30
Fix optimizers to init symbolVariable as late as possible

steffengy:
Small example: (pseudo-code)
``
let val = OPTIMIZED_FUNC(val);
``
can result in
```
ZEPHIR_INIT_NVAR(_1); //NOTICE that here the previous value of _1 get's destroyed
ZVAL_DOUBLE(val, OPTIMIZED_FUNC_C(zephir_get_numberval(_1)));
```

Behavior can be that the result is simply 0, since zephir_get_numberval receives
an "empty" zval.
Therefor the best approach is to delay the initVariant atleast until the code
for parameter resolving has been generated.

### DIFF
--- a/Library/Optimizers/FunctionCall/AddcslashesOptimizer.php
+++ b/Library/Optimizers/FunctionCall/AddcslashesOptimizer.php
@@ -61,16 +61,15 @@ class AddcslashesOptimizer extends OptimizerAbstract
             throw new CompilerException("Returned values by functions can only be assigned to variant variables", $expression);
         }
 
-        if ($call->mustInitSymbolVariable()) {
-            $symbolVariable->initVariant($context);
-        }
-
         $context->headersManager->add('kernel/string');
 
         $symbolVariable->setDynamicTypes('string');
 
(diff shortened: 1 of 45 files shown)


andresgutierrez:
Could you please provide the code that produces the issue in first instance?

steffengy:
Something like
``
var g = 5;
        let g = round(g / 2);
``
should be enough to trigger it.

``
	zval *g = NULL, _0;

	ZEPHIR_MM_GROW();
	ZEPHIR_INIT_VAR(g);
	ZVAL_LONG(g, 5);

	ZEPHIR_INIT_NVAR(g);
	ZEPHIR_SINIT_VAR(_0);
	ZVAL_DOUBLE(&_0, zephir_safe_div_zval_long(g, 2 TSRMLS_CC));
	zephir_round(g, &_0, NULL, NULL TSRMLS_CC);
	ZEPHIR_MM_RESTORE();
``

ovr:
Nice :+1:

PR #977, opened 2015-06-01
Improve api generation

gsouf:
In this PR i added some stuff i needed for the phalcon 2 api generation : 

- a command line argument parser (copyed from  https://github.com/pwfisher/CommandLine.php)
- a fix for the way the version was got in the default api theme, additionally we now have access to the config from a template
- 2 ways to get a theme : 
  - from the command line ``zephir api theme-path=/path-to-theme``
  - from the config by adding theme directories (``$config["api"]["theme-directories"]``)

I'll try to add the documentation asap.

Let me know if something is wrong

### DIFF
--- a/Library/CommandArgumentParser.php
+++ b/Library/CommandArgumentParser.php
@@ -1,0 +1,207 @@
+<?php
+
+/*
+ +----------------------------------------------------------------------+
+ | Zephir Language                                                      |
+ +----------------------------------------------------------------------+
+ | Copyright (c) 2013-2015 Zephir Team                                  |
+ +----------------------------------------------------------------------+
+ | This source file is subject to version 1.0 of the MIT license,       |
+ | that is bundled with this package in the file LICENSE, and is        |
+ | available through the world-wide-web at the following url:           |
+ | http://www.zephir-lang.com/license                                   |
+ |                                                                      |
+ | If you did not receive a copy of the MIT license and are unable      |
+ | to obtain it through the world-wide-web, please send a note to       |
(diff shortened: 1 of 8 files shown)


gsouf:
@phalcon @ovr an you review it please ?

ogarbe:
In my opinion you shouldn't add the command line argument parser (copyed from https://github.com/pwfisher/CommandLine.php) but use composer to add it to the project.

gsouf:
@ogarbe since the begining of zephir we want to avoid using composer or external dependencies. Instead we want to package everything in zephir itself. 

That's why we even have written custom code for template generation instead of using some twig, or other template engine

ogarbe:
i don't understand why but ... so be it...
Bitwise if i look in zephir there is some composer requirements of external dependencies : https://github.com/phalcon/zephir/blob/master/composer.json#L20-L32

gsouf:
The goal is to make zephir available as a standalone.

Indeed there are composer depencies. The composer depencies are : 
- in ``require`` : the extensions that zephir needs to run. However composer cant install extension, it's only to make the composer.json correct with real dependencies for people that install it with composer.
- in ``require-dev`` : it's the dependencies needed when we want to develop and contribute to zephir.

ogarbe:
i don't get why using composer for external dependencies prevent to be a standalone (creating a phar with all libraries + vendors would work completely), but if it's the desired design, it's ok :-)

gsouf:
Only @phalcon will give you the reasons why, I think the most important is to keep things clear !

andresgutierrez:
Thanks

PR #978, opened 2015-06-01
Fix #537: Make sure class properties are initialized in the right order

### LINKED ISSUE #537: Can't overwrite the default array data if extends [$10 awarded]
I can't set the default array data if extends:

``` zephir
class Test
{

    protected _data = [] { set, get };
}
```

``` php
<?php

class Users extends Test
{

    protected $_data = [
        'id', 
        'username', 
        'password', 
    ];
}
```

``` php
$user = new Users();
var_dump($user->getData())
```

> array(0) { }

Issue related to #520

steffengy:
Testcase from #537 adopted.

Previously the constructor code of the parent class was appended,
because the parents constructor method was simply reused.
This explicitly prepends only the statements.

### DIFF
--- a/Library/ClassProperty.php
+++ b/Library/ClassProperty.php
@@ -255,7 +255,17 @@ class ClassProperty
                     $visibility = array('internal');
                 }
 
+                $constructParentMethod = null;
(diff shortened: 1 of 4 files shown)


ovr:
Thank you
Awesome work :cat:

Issue #982, opened 2015-06-02
how to get config from php.ini?


shanghaiyangming:
How to get config from php.ini?

like this 

my_setting.a = "tcp://127.0.0.1"

steffengy:
should work just as in PHP: http://php.net/manual/de/function.ini-get.php

shanghaiyangming:
i do this things, but it not work
let this->_expire = ini_get("max_execution_time"); // it's ok
let config = ini_get("my.utils.lock.memcached");// it's bad

andresgutierrez:
Implemented in #985

Issue #660, opened 2014-12-01
Add support for declaring ini settings entries


wapmorgan:
Like `globals` section in **config.json** there can be section `ini-entries`:

```
  'ini-entries': {
    'available-operators': {
      'type': 'bool or int, or real, or string, or unempty-string'
      'permission': 'system or perdir, or user, or all'
    }
  }
```

[1] http://docstore.mik.ua/orelly/weblinux2/php/ch14_12.htm
##

PR #985, opened 2015-06-04
Globals setup in php.ini

ogarbe:
Now if you have global defined in config.json, all bool value can be setup using php.ini

tested with phalcon 2.0 with orm.xxx parameters

can also help for https://github.com/phalcon/zephir/issues/982

can also be used for https://github.com/phalcon/zephir/issues/660 (i don't see the use of separating global and ini-entries)


Library/Code/Builder/Struct.php
```diff
@@ -139,11 +139,12 @@ public function getCDefault($name, $global)
 
             case 'boolean':
             case 'bool':
+return '';
```
ogarbe: pretty bad but didn't manage not to overwrite value otherwise

ogarbe:
@andresgutierrez what do you think of this ?

andresgutierrez:
Could you please rebase? What would happen if there is a default value in config.json?

ogarbe:
the default value is used from config.json, and can be override in php.ini

andresgutierrez:
Thanks

leo108:
Is it possible to support string value?

PR #987, opened 2015-06-05
Fix ugly bug inside instanceof

baszczewski:
This function doesn't work properly before change, when compared class name have same length but other content. Now part with Z_OBJCE_P is used only when we compare with same class which is ofcourse faster than zend_fetch_class. Example: 

var a;
let a = new Test();
var_dump(a instanceof Test);


andresgutierrez:
Thanks

PR #991, opened 2015-06-09
API generation - Fix theme naming bug and add output directory to command arg

gsouf:
This is some things needed for phalcon 2 api generation.

I will maybe add more commits, feel free to merge or not, i can open a new PR if you do


andresgutierrez:
Thanks

PR #996, opened 2015-06-13
Fix #994: Make sure to also check the elseif statements for mutations

### LINKED ISSUE #994: Not called constructor of the intended class.
Not called constructor of the intended class.

It occurs when you use the elseif clause.

Environment:

* PHP 5.6.9
* Zephir 0.7.0a (b190e04)

zephir code: test/loader.zep:

```
namespace Test;

class Loader
{
    public static function t1(array classes)
    {
        var key, name, instances;

        let instances = [];

        for key, name in classes {
            if typeof name == "object" {
                let instances[key] = name;
            } else {
                if typeof name == "string" {
                    var instance;
                    let instance = new {name}();
                    let instances[key] = instance;
                } else {
                    throw new \Exception("Invalid class object");
                }
            }
        }

        return instances;
    }

    public static function t2(array classes)
    {
        var key, name, instances;

        let instances = [];

        for key, name in classes {
            if typeof name == "object" {
                let instances[key] = name;
            } elseif typeof name == "string" {
                var instance;
                let instance = new {name}();
                let instances[key] = instance;
            } else {
                throw new \Exception("Invalid class object");
            }
        }

        return instances;
    }
}
```

test code: `test.php`

```
<?php
namespace Test;

class Hoge {
    public function __construct() {
        echo __METHOD__, PHP_EOL;
    }
}

class Foo {
    public function __construct() {
        echo __METHOD__, PHP_EOL;
    }
}

if (isset($argv[1])) {
    echo "-- t2\n";
    Loader::t2([
        '\\Test\\Hoge',
        '\\Test\\Foo',
    ]);
} else {
    echo "-- t1\n";
    Loader::t1([
        '\\Test\\Hoge',
        '\\Test\\Foo',
    ]);
}
```

execute result:

```
$ php a.php
-- t1
Test\Hoge::__construct
Test\Foo::__construct
$ php a.php t2
-- t2
Test\Hoge::__construct
Test\Hoge::__construct  # Called Test\Hoge::__construct (Not Test\Foo)
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/21618464-not-called-constructor-of-the-intended-class?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github).
</bountysource-plugin>

steffengy:
This was more trivial than expected:

A method cache was used leading to the first constructor being called
multiple times, since elseif branches weren't checked for mutations,
leading to the usage of caches.

### DIFF
--- a/Library/Passes/MutateGathererPass.php
+++ b/Library/Passes/MutateGathererPass.php
@@ -299,6 +299,9 @@ class MutateGathererPass
                     if (isset($statement['else_statements'])) {
                         $this->passStatementBlock($statement['else_statements']);
                     }
+                    if (isset($statement['elseif_statements'])) {
+                        $this->passStatementBlock($statement['elseif_statements']);
+                    }
                     break;
 
                 case 'switch':


andresgutierrez:
Thanks

PR #1005, opened 2015-06-15
Rework property initialization to use create_object insteadof constructors

steffengy:
Another step towards consistent constructor (property initialization) behavior:

Effective changes:
------------
- Ensures that classes are compiled in the right dependency order
- The parent constructor generation step is not necessary anymore, since
  property initialization is now in create_object of each class. (before the constructor)
- Allows php defaults (defaults of a PHP class which extends a ZEPHIR class) to overwrite it (except null, which would be tricky with the current
implementation)
- Does remove parent initialization logic if it's anyways overwritten by child initialization
(e.g. parent initializes property ``data`` to an empty array, which then is overwritten to an array containing a string resulted in 2 property assignments, which now results in 1)

### DIFF
--- a/Library/ClassDefinition.php
+++ b/Library/ClassDefinition.php
@@ -865,7 +865,7 @@ class ClassDefinition
     private function getInternalSignature(ClassMethod $method)
     {
         if ($method->getName() == 'zephir_init_properties') {
-            return 'static void ' . $method->getName() . '(zval *this_ptr TSRMLS_DC)';
+            return 'static zend_object_value ' . $method->getName() . '(zend_class_entry *class_type TSRMLS_DC)';
         }
         if ($method->getName() == 'zephir_init_static_properties') {
             $classDefinition = $method->getClassDefinition();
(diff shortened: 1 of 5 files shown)


andresgutierrez:
"chaos"

steffengy:
@andresgutierrez
reworded that, since it didn't really describe the problem as accurately as it should.

andresgutierrez:
Nice, it seems the build is failing because of slowness in a Travis box, it looks good so far

steffengy:
@andresgutierrez 
Yeah the failing travis tests are strange,
waiting for feedback from @mruz.

Hopefully there is not more broken in that area^^

andresgutierrez:
Did you finish the work on this?

steffengy:
@andresgutierrez 
Yeah seems to work fine

andresgutierrez:
Thanks!

andresgutierrez:
After this commit, compilation is taking a lot of memory, do you have any idea?

```php
> zephir build
PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 130968 bytes) in /Users/andresgutierrez/zephir/Library/ClassMethod.php on line 1138
```

steffengy:
That is strange, haven`t noticed something like that.
I`ll look into it as soon as possible

andresgutierrez:
It looks it's caused by a recursive calculateDependencies() with circular references.

andresgutierrez:
I found the problem, as now the classes are compiled in a different order, class constants were not compiled when expected causing cyclic compilations.

PR #1006, opened 2015-06-15
Optimize zephir_require to work with opcache

steffengy:
If you have opcache active and require something from within zephir,
it isn't as efficient as it could be:

it checks and opens the file again, even if it wouldn't have to, since
the op array is already cached.

This implementation is more like the PHP-internal approach (zend_vm_def.h / compile_filename),
which fixes this unnecessary "overhead".


Green-Cat:
:+1:

steffengy:
@andresgutierrez 
could you review this?

andresgutierrez:
Thanks

PR #1008, opened 2015-06-16
Fixed ClassMethod::getNumberOfRequiredParameters

kjdev:
methods of built-in classes becomes error in check of argument.

ex: test.zep

```
use Phalcon\Mvc\View\Engine;
use Phalcon\Mvc\View\EngineInterface;

class Test extends Engine implements EngineInterface
{
    public function render(string path, var params, boolean mustClean = false)
    { ... }
}
```

execute enerate.

```
$ zephier generate
Zephir\CompilerException: Class Test::partial() does not have the same number of required parameters in interface: Phalcon\Mvc\View\Phalcon\Mvc\View\EngineInterface
```

Phalcon\Mvc\View\EngineInterface

```
public function partial(string! partialPath) -> string;
```

Phalcon\Mvc\View\Engine

```
public function partial(string! partialPath, var params = null) -> string
{ ... } 
```

`ClassDefinition::checkInterfaceImplements`

```
$implementedMethod->getNumberOfRequiredParameters() => 2  # or not properly 1
$method->getNumberOfRequiredParameters() => 1
```

### DIFF
--- a/Library/ClassMethod.php
+++ b/Library/ClassMethod.php
@@ -577,9 +577,10 @@ class ClassMethod
             if (count($parameters)) {
                 $required = 0;
                 foreach ($parameters as $parameter) {
-                    if (!isset($parameter['default'])) {
-                        $required++;
+                    if (isset($parameter['default']) || (isset($parameter['mandatory']) && !$parameter['mandatory'])) {
+                        continue;
                     }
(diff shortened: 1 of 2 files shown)


kjdev:
It will cancel because the implementation was wrong.

PR #1014, opened 2015-06-22
A theme can now extend an other theme

gsouf:
A theme can extend an other, that allows to use an existing and only have to change some css or minor parts


andresgutierrez:
Nice

Issue #952, opened 2015-05-15
PHP7


andresgutierrez:
It seems the generation of extensions from PHP5 changes dramatically in PHP7 due to the modifications to internal structures.

As I see it will require:

* A separate kernel specifically focused on PHP7
* A new memory manager designed to address the reference counting changes in PHP7
* A new approach on generating method/function calls
* A new approach on reusing zval variables
* It would be hard to have a common generated code for both PHP5 and PHP7

Also, It seems that the amount of changes is larger than I expected previously and I think these changes will take many hours and it's something we have to address as soon as possible before move forward.

andresgutierrez:
Old issue https://github.com/phalcon/zephir/issues/756

steffengy:
@andresgutierrez 
> A separate kernel specifically focused on PHP7
> A new memory manager designed to address the reference counting changes in PHP7
> A new approach on reusing zval variables
> It would be hard to have a common generated code for both PHP5 and PHP7

Totally agree on this points, as that matches the result of experiments I conducted.

> A new approach on generating method/function calls

This actually has not changed very much, what problem do you see here?

I think another issue will be that currently we have localOnly and "global" (heap) zval's,
which result in different handling on the compiler side (PHP).
Any idea how to implement the different code generation for PHP7?
It seems like it will require huge amounts of changes (and removing some duplicated code)
to achieve that. (And therefore could easily result in a mess with very much if's)

Another interesting example for that is basically the StringsManager, which creates C code for concatenation, which also has to be changed for PHP7 (zend_string, no ``zval **`).

Some examples of kernel functions can be taken from my first/second experiment,
while it should only be used as reference, since some implementations are kept
too simple to not consume all my time by porting that complexity 
(e.g. array multi update in first experiment).

I still think the most important part will be to find a good way to implement the code generation
for PHP5/PHP7 in the compiler, which will require a smart way to solve it, since you basically
can not leave many places untouched...

hylent:
:+1:

baszczewski:
In my opinion you can abandon support to old php version and make more efforts to support PHP 7. Everyone who will use Zephir, probably can install new php version to.

steffengy:
@andresgutierrez 
What do you think about the approach in 12bbcba4a014f90c664498c6ce57a25fe5500c2d

Basically it's a separate bare kernel, which is used when compiling using PHP7, 
with a basic "skeleton" needed to compile an empty extension,
which should allow porting feature by feature.

[...]

PR #1015, opened 2015-06-22
Initial preparation for refactoring and moving forward

steffengy:
Changes:
- Template/Kernel files dependant on backend
- Active backend detection (PHP5->ZE2,PHP7->ZE3), 
- if no backend is specified via --backend ZendEngine2 for example
- Bare kernel for ZE3 and basic templates

This allows moving forward with PHP7 support,
because part by part implementation and refactoring
code generation in the process of it is made possible,
without breaking BC.


steffengy:
ping @andresgutierrez 
Here you go, the PR you wanted.
the travis fail is due to a build timeout (>10 mins, whyever they seem to happen now...)

steffengy:
@andresgutierrez 
Can you take a look at this?
It's a direct PR of what you wanted as such in #952

andresgutierrez:
@steffengy Looking at this right now :)

steffengy:
@andresgutierrez
Did you already take a look at it? :)

andresgutierrez:
@steffengy Do you have more incoming changes?

steffengy:
@andresgutierrez 
I am working on refactoring the parts required for PHP7 compatibility.

Basically introduce support feature by feature and refactoring part by part 
until 100% is reached, which is what I'm working on at a separate branch
( https://github.com/steffengy/zephir/tree/ng-refactor2 )

I think it makes sense to first merge this as a preparation.
This also isn't very dangerous in terms of BC.

Rolling in new changes as they are stable in regards of BC
(currently that other tree is failing compilation on 5.X, because it's very heavily in
development and in the progress stuff during that unfortunately sometimes breaks)
makes probably more sense and should be handled as a/multiple separate PRs.

steffengy:
To clarify: 
That means no there won't be any more changes regarding that in **this PR**
and I would like this merged, if there isn't any problem.

andresgutierrez:
Thanks!

Issue #1011, opened 2015-06-17
Constant as other const


mruz:
> Zephir\CompilerException: Type "static-constant-access" is not supported.

```zep
class Version
{

    const DEV = 0;
    const ALPHA = 1;
    const BETA = 2;
    const RC = 3;
    const STABLE = 4;

    const MAJOR = 1;
    const MINOR = 0;
    const PATCH = 13;
    const STAGE = self::STABLE;
    const BUILD = 1;
    
    public static function getStage() {
        return self::STAGE;
    }
}
```
```php
echo Version::getStage(); //4
```

mruz:
Also it could be nice to can acces to const by static https://github.com/phalcon/cphalcon/pull/10515
```zep
class A
{
    const VERSION = 1;
    
    public static function getVersion()
    {
        return static::VERSION;
    }
}

class B extends A
{
    const VERSION = 2;
}
```
```php
var_dump(B::getVersion()); //2
```

@ogarbe is this related to https://github.com/phalcon/zephir/issues/513?

hylent:
:+1:

ovr:
Seems to be fixed

mruz:
```zep
class Version
{

    const DEV = 0;
    const ALPHA = 1;
    const BETA = 2;
    const RC = 3;
    const STABLE = 4;

    const MAJOR = 1;
    const MINOR = 0;
    const PATCH = 14;
    const STAGE = self::STABLE;
    const BUILD = 1;
```

> Exception: Cannot resolve constant

If I replace by class name:
```zep
    const STAGE = Version::STABLE;
```
> PHP Warning:  require(/home/mruz/Dropbox/git/ice/framework/vendor/phalcon/zephir/LibraryVersion.php): failed to open stream: No such file or directory in /home/mruz/Dropbox/git/ice/framework/vendor/phalcon/zephir/Library/Loader.php on line 48
PHP Fatal error:  require(): Failed opening required '/home/mruz/Dropbox/git/ice/framework/vendor/phalcon/zephir/LibraryVersion.php' (include_path='.:/usr/share/php5:/usr/share/php5/PEAR') in /home/mruz/Dropbox/git/ice/framework/vendor/phalcon/zephir/Library/Loader.php on line 48

ovr:
> Exception: Cannot resolve constant


@mruz 
It's not possible to get this error, Did you update Zephir to latest master?


I tested 

```zep
//works
const STAGE = self::STABLE;
//works
[...]

PR #1017, opened 2015-06-26
Implement #1011 issue - static-constant-expression for ClassConstants

ovr:


steffengy:
Im not quite sure if it makes sense why not to squash some of these fix psr commits...

ovr:
@steffengy 
I was working on different PCs, and It's takes more than 1 hour....
And I reviewed it per commit via 155 files changed :)

PR #1027, opened 2015-07-10
CI: Migrate travis to new container based infrastructure

steffengy:
This allows faster and more consistent build times,
and potentially new features

More information via
http://docs.travis-ci.com/user/migrating-from-legacy/

PR #1031, opened 2015-07-15
Removed sudo from json-c installer

SidRoberts:
Related to #1027.

### DIFF
--- a/install-json
+++ b/install-json
@@ -3,4 +3,4 @@ git submodule update --init
 cd json-c
 sh autogen.sh
 ./configure
-make && sudo make install
+make && make install


steffengy:
What is the reasoning behind this?

I cant really see the sense here, because you need the root permissions,
to install in /usr/local/lib/ , which is what this very likely tries to do.

The only reason it works for the travis parts is because they use a separate prefix,
to prevent this permission conflict.

SidRoberts:
Phalcon no longer passes in Travis is due to [json-c not being installed properly](https://travis-ci.org/phalcon/cphalcon/jobs/71008323#L826) and the only thing I can think of is that Travis blocks `sudo` commands under the new infrastructure - meaning [`sudo make install`](https://github.com/phalcon/zephir/blob/master/install-json#L6) won't run.

Should we force people to run `sudo ./install-json` or create `./install-json-nosudo`?

steffengy:
Json-C isn't the problem there, since it's installed to /home/travis/build/phalcon/cphalcon/build/lib/.

SidRoberts:
It is the problem, see https://github.com/phalcon/cphalcon/pull/10667.

steffengy:
It is not, the problem is simply that phalcon had the wrong lookup paths for libraries (as you also see due to the nature of the error message 
``parser.c:9:25: fatal error: json-c/json.h: No such file or directory``),
which is fixed by either setting the environment variables mentioned above or by doing the
fix applied: d6f25fcaef3d46c03ebf2a7ee4de2d2062286a16

It is fixed upstream (not the way I would fix it but fixed), as you can also see as phalcon's travis tests pass.

SidRoberts:
Fair enough. :smile:

PR #1047, opened 2015-08-06
Fixes docblock

Green-Cat:
getClassDefinition can return false if a class definition is not found.


andresgutierrez:
Could you please submit this to the development branch?

Green-Cat:
Done, see #1064

PR #1048, opened 2015-08-06
Fixes loose comparison

Green-Cat:
Strict comparison should be used for booleans,
If this is approved I can make a PR with all the other bool comparisons in the project.


Green-Cat:
Resubmited against dev #1068.

PR #1050, opened 2015-08-06
branch property is missing

Green-Cat:


andresgutierrez:
Could you please submit this to the development branch?

Green-Cat:
Done, see #1063

ovr:
@andresgutierrez It's not needed,
I suggest to use development for BC changes like main future PHP 7

PR #1051, opened 2015-08-06
Removes unneeded variable declaration.

Green-Cat:


Green-Cat:
Resubmited against dev #1069.

PR #1053, opened 2015-08-06
PHP7: Fix some bugs, generally add more compatibility

steffengy:
More work towards  the development branch.

Just merge it after about a week when there are several commits piled up,
so i don't have to create loads of PRs

### DIFF
--- a/Library/Backends/ZendEngine2/Backend.php
+++ b/Library/Backends/ZendEngine2/Backend.php
@@ -980,7 +980,7 @@ class Backend extends BaseBackend
     {
         $paramStr = $params != null ? ', ' . join(', ', $params) : '';
         if (!isset($symbolVariable)) {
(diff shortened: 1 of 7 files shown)


andresgutierrez:
Thanks. I'm wondering if we should freeze master branch

steffengy:
@andresgutierrez 
That would atleast reduce the amount of possible merge conflicts, right?

ovr:
It's not needed to freeze master,
For example I can merge master to develop....

ovr:
@andresgutierrez 
Maybe it's time to give write access to @steffengy ?
He can works in `development` branch

sergeyklay:
@ovr +1

steffengy:
@ovr
Of course it's not needed and works without, but the main reasons I think would be to
- keep master stable
- reduce the amount of merges, by introducing all changes in development (insteadof master)
  and then merging one time, once it's deemed stable

I also wouldn't mind write access, but that is also no requirement :)

ovr:
@steffengy 
Write access is needed
You can push PHP 7 changes directly to `development branch`

steffengy:
@ovr 
I agree that if I had those rights, it would make some stuff much easier.
But until then I guess I'll have to create some PRs regularily.

Issue #1055, opened 2015-08-15
str_replace null error BUG


imadige:
let str2=str_replace(["aa"],["aaa"],"bla bla aa aaa askks");

str2 null

imadige:
http://php.net/manual/en/function.str-replace.php


<?php
// Provides: <body text='black'>
$bodytag = str_replace("%body%", "black", "<body text='%body%'>");

// Provides: Hll Wrld f PHP
$vowels = array("a", "e", "i", "o", "u", "A", "E", "I", "O", "U");
$onlyconsonants = str_replace($vowels, "", "Hello World of PHP");

// Provides: You should eat pizza, beer, and ice cream every day
$phrase  = "You should eat fruits, vegetables, and fiber every day.";
$healthy = array("fruits", "vegetables", "fiber");
$yummy   = array("pizza", "beer", "ice cream");

$newphrase = str_replace($healthy, $yummy, $phrase);

// Provides: 2
$str = str_replace("ll", "", "good golly miss molly!", $count);
echo $count;
?>

sergeyklay:
Dublicate https://github.com/phalcon/zephir/issues/732

lolo32:
I post a merge request #1059 to solve this problem.

steffengy:
Fixed in development.

PR #1059, opened 2015-08-18
Correct str_replace using first argument an array

lolo32:
Since v0.7.0 (work with this one), it's not possible to use the function

``` php
str_replace(Array, Array|String, String)
```

in zephir (see bug [#1055] for the latest one). If the first argument is an array in the function str_replace, the return was always NULL. The cause is the commit [c763915fc619151ca5498c9377f8df0da51f5536], solving a memory leak. I think, this is a C pointer error. This work correctly for me, in master after doing this modification).


ovr:
Can you push a test for it?

sergeyklay:
@lolo32

steffengy:
Fixed in development.

PR #1067, opened 2015-08-25
Improves performance of the getResolveParamsAsExp method

Green-Cat:
resubmit of #1046.
Due to the way for loops are implemented in php, the count function will run every iteration of the loop. Therefore, moving the count to a variable, calculated before the start of the loop, improves performance (less function calls).


steffengy:
I'm not quite sure if it makes sense to make micro-optimizations in that area 
(improvement of 0,5 ms on 10k iterations), but I'll merge this anyways.

PR #1074, opened 2015-08-30
Implement internal function optimization, allow bypassing php userspace

steffengy:
This generates 2 versions of methods, a PHP exposed definition and an
internal version of the method.
(default: disabled, enable using the config option `internal-call-transformation`)
During a zephir method call, whenever possible, 
the internal version is prefered, leading to better performance.

This implementation also tries to preserve the appropriate scope, to allow parent::
calls and stuff depending on the scope to work properlywithin internal methods.
## Limitations
- Not applicable for dynamic or `static::` calls
## Todos:
- [x] Implement for ZendEngine2 (seems more unstable than for ZE3)
- [x] Cleanup


Green-Cat:
:+1:

sergeyklay:
:+1:

PR #1075, opened 2015-09-01
Fixes loose comparisons

Green-Cat:
Ref #1068


andresgutierrez:
In which case these are needed?

Green-Cat:
It is just a good practice, since a loose comparison (==) will match 0 to false and any positive integer or string to true. It can create bugs which are really hard to detect in cases where (for example) false is passed as a string or another similar small error.

PR #1073, opened 2015-08-26
Prove bug #1062

fezfez:

Issue #1062, opened 2015-08-21
Fatal error at compilation when an extended/implemented class does not exist


fezfez:
Hi,

I have this error

```
Argument 1 passed to Zephir\ClassDefinition::setExtendsClassDefinition() must be an instance of Zephir\ClassDefinition, instance of Zephir\ClassDefinitionRuntime given, called in /home/travis/build/fezfez/php-to-zephir/vendor/phalcon/zephir/Library/CompilerFile.php on line 755 and defined
/home/travis/build/fezfez/php-to-zephir/vendor/phalcon/zephir/Library/ClassDefinition.php:385
/home/travis/build/fezfez/php-to-zephir/vendor/phalcon/zephir/Library/CompilerFile.php:755
/home/travis/build/fezfez/php-to-zephir/vendor/phalcon/zephir/Library/Compiler.php:765
/home/travis/build/fezfez/php-to-zephir/vendor/phalcon/zephir/Library/Commands/CommandAbstract.php:101
/home/travis/build/fezfez/php-to-zephir/src/PhpToZephir/CodeValidator.php:28
/home/travis/build/fezfez/php-to-zephir/tests/Converter/ConverterBaseTest.php:91
/home/travis/build/fezfez/php-to-zephir/tests/Converter/Code/Oop/ExtendsAndImplementTest.php:134
```

I've found the errored line in 0.7.1

https://github.com/phalcon/zephir/blob/182c87d35201a8d4cb4c44c27955f325a1f73b43/Library/CompilerFile.php#L755
https://github.com/phalcon/zephir/blob/182c87d35201a8d4cb4c44c27955f325a1f73b43/Library/CompilerFile.php#L722

You can see the error in my travis build :

- Error on Implements : https://travis-ci.org/fezfez/php-to-zephir/jobs/76591386#L270
- Error on Extends : https://travis-ci.org/fezfez/php-to-zephir/jobs/76591386#L259

steffengy:
Could you try using the development branch?

lajosbencz:
I've just stumbled upon the same issue today, updated to the development branch and now compilation succeeds!

fezfez:
Using the dev branch throw an exception.

```
Zephir\CompilerException: Cannot locate interface Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface when implementing interfaces on Symfony\Component\EventDispatcher\DependencyInjection\RegisterListenersPass
at Library/ClassDefinition.php(1104)

```

[...]

PR #1076, opened 2015-09-01
Prove bug #1062

fezfez:
See #1073


fezfez:
@andresgutierrez @steffengy : Any chance to merge this PR ? :beers:

steffengy:
Should be fine to merge.

I was a bit sceptical at first, since you moved the aliasManager initialization
to the constructor, which itself is no problem.

But if the alias manager would contain information that's different per compilation,
you wouldn't be able to call preCompile multiple times (if that's ever needed)
on a compilation file. 

Since alias managing should not be able to produce conflicts of any kinds, on multiple
runs, I'll merge this.

PR #1077, opened 2015-09-01
Activate code coverage with scrutinizer-ci

fezfez:


fezfez:
Ooops.. create PR in master...

Issue #897, opened 2015-04-19
[BUG] Interface constant


KorsaR-ZN:
see https://github.com/zephir-lang/zephir/issues/5

PR #1093, opened 2015-09-24
Fixes bug not resolve constant from interface

KorsaR-ZN:
see #897


steffengy:
@KorsaR-ZN
Seems fine on the first look.
Could you provide a small test case, to show that the bug is resolved and prevent it from reappearing?

KorsaR-ZN:
@steffengy added tests

steffengy:
Thank you!

sergeyklay:
:+1:

PR #1101, opened 2015-09-29
Added new expression builder and also min-refactoring

KorsaR-ZN:
Please merge


steffengy:
That looks great, will review it later.

PR #1104, opened 2015-09-30
[WIP] First implementation of zept

fezfez:
Please dont merge now
- [X] Implement first test case
- [ ] Convert .phpt test from php-src into .zept
- [ ] Unit test on zephir-testcase
- [ ] Give zephir-testcase to Zephir organisation ?

Comments are welcome !


steffengy:
If that's implemented, regarding the current (not modular) architecture, it shouldn't be as an external component but directly implemented into zephir, effectively superseding the old tests.

PR #1107, opened 2015-10-02
Initialize parent array properties, fix #1033

### LINKED ISSUE #1033: CompilerException: Class does not have a property
With the latest changes I'm receiving exception:

> mruz@linux-9cy2:~/Dropbox/git/ice/framework> ./vendor/bin/zephir build
Zephir\CompilerException: Class 'Ice\Config' does not have a property called: 'messages'

It doesn't has, but previously worked, and I don't know where the problem is.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24429004-compilerexception-class-does-not-have-a-property?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F280146&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mruz:


steffengy:
Nice catch

PR #1108, opened 2015-10-02
Bug seen while reading code

joeyhub:
I don't know if the bad end_address pointer affects anything or if it is used at all but if it is this should set it to the correct address rather somewhere randomly in the stack.

### DIFF
--- a/ext/kernel/memory.c
+++ b/ext/kernel/memory.c
@@ -386,7 +386,7 @@ void zephir_initialize_memory(zend_zephir_globals_def *zephir_globals_ptr TSRMLS
 	}
 
 	zephir_globals_ptr->start_memory = start;
-	zephir_globals_ptr->end_memory   = start + ZEPHIR_NUM_PREALLOCATED_FRAMES;
+	zephir_globals_ptr->end_memory   = &start[ZEPHIR_NUM_PREALLOCATED_FRAMES - 1];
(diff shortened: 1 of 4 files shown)


steffengy:
5 commits with 1 line change each?
That really should be squashed.

joeyhub:
Sorry about that. I was being a bit lazy and edited them in github. I didn't plan to make changes so quickly when I started examining the code so never checked out. Though practice wise if there was more de-duplication it might be less of an issue. I'm not sure why the same routine is duplicated in so many places. I'm not sure how to squash but it can't be that difficult, if I find the time. If it looks like I'm going to make more fixes I'll checkout a copy.

joeyhub:
That was relatively painless. Enjoy.

sergeyklay:
@joeyhub Could you please rebase

sergeyklay:
@joeyhub

sjinks:
I don't think that the fix is correct. start_memory and end_memory were the bounds for the preallocated memory frames. Preallocated memory frames persist between requests and were used to improve performance (memory allocation is not cheap).

```c
int persistent = (frame >= g->start_memory && frame < g->end_memory);
void *buf = perealloc(frame->addresses, sizeof(zval **) * (frame->capacity + 16), persistent);
```

Issue #1099, opened 2015-09-27
Introduce proper variable scoping


steffengy:
``` zep
class Test1
{
    /**
     * Output: int(6) int(7) int(8) int(9) int(10)
     * Expected: int(6) int(6) int(6) int(6) int(6)
     */
    public function testA()
    {
        var k, v;
        //var s = this->testB(); should work by creating an implicit let

        for k, v in range(1, 5) {
/* Scope for variable "g" begins */       
            var g = 5;
            let g++;
            var_dump(g);
/* Scope for variable "g" ends */
        }
        var_dump(g); // currently int(10), This shouldn't compile!
        // but throw an error "Variable is not declared in the current scope"
    }

   public function testB()
   {
        var b = "test"; //variable rootscope
        if b == "test" {
/* scope for new local variable "b" */
              var b = 5; //Parent scope variable is shadowed, a new variable is introduced for that scope
              var_dump(b); //Expect: int(5)
/* end of scope for variable "b" */
        }
        var_dump(b); //Expect: string("test")
   }
}
```

andresgutierrez:
+1

sergeyklay:
+1

KorsaR-ZN:
:+1:

steffengy:
Initial support available through https://github.com/phalcon/zephir/pull/1109 .

steffengy:
should be available in master.

PR #1109, opened 2015-10-03
Refactor variable declaration, initial support for scopes #1099

steffengy:
Basic implementation, experimental and BC-break.


andresgutierrez:
good job!

kse300489:
Ty, nice feature

Issue #1058, opened 2015-08-16
[NFR] ldexp


sergeyklay:
PHP & Zephir does not have `ldexp(x,n)` function from Math library:
http://www.codecogs.com/library/computing/c/math.h/ldexp.php

> The *ldexp* functions multiply x by 2 to the power of n.

**go**
```go
math.Ldexp(2,3) // 16
```
**ruby**
```ruby
Math.ldexp(2,3) # 16.0
```

**python**
```python
math.ldexp(2,3) # 16.0
```

**c**
```c
#include <math.h>
ldexp(2.0,3) // 16.0
```

It may be implemented like that:
```zephit
function ldexp(x, n)
{
   return x * pow(2, n);
}
```

ovr:
Will do it tomorrow

sergeyklay:
:+1:

ovr:
Started, I am working on this problem now
Will create optimizer for Zephir
And NFR for PHP core

ovr:
refs https://github.com/php/php-src/pull/1489

sergeyklay:
any progress?

sergeyklay:
Done

PR #1128, opened 2015-11-16
Fixed issue #1058

sergeyklay:


sergeyklay:
The function
```zephir
ldexp(double x, int exponent) -> double
```
returns `x` multiplied by 2 raised to the power of `exponent`.

PR #1129, opened 2015-11-17
Fixed issue #1058

sergeyklay:
#### Description

The function **ldexp(double x, int exponent)** returns **x** multiplied by 2 raised to the power of **exponent**.
#### Declaration

Following is the declaration for ldexp() function.

``` zephir
ldexp(double x, int exponent) -> double
```

Zephir uses type casting.
#### Parameters
- **x** − This is the floating point value representing the significand.
- **exponent** − This is the value of the exponent.
#### Return Value

This function returns x \* 2 exp
#### Example

The following example shows the usage of ldexp() function.

**Zephir**

``` zephir
namespace Utils;

class Test
{
    public function foo(var x, var exponent)
    {
        return ldexp(x, exponent);
    }
}
```

**PHP**

``` php
use Utils\Test;

$test = new Test;
echo $test->foo(2, 3); //  16.0
```

### DIFF
--- a/Library/Optimizers/FunctionCall/LdexpOptimizer.php
+++ b/Library/Optimizers/FunctionCall/LdexpOptimizer.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ +--------------------------------------------------------------------------+
+ | Zephir Language                                                          |
+ +--------------------------------------------------------------------------+
+ | Copyright (c) 2013-2015 Zephir Team and contributors                     |
+ +--------------------------------------------------------------------------+
+ | This source file is subject the MIT license, that is bundled with        |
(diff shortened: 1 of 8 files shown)


sergeyklay:
The function
```zephir
ldexp(double x, int exponent) -> double
```
returns `x` multiplied by 2 raised to the power of `exponent`.

steffengy:
@sergeyklay 
you'll have to rebase against master, then the tests should pass.
also it might be good to change the commit message to something more verbose (since you can't guess what actually was changed by just looking at the history, without looking up the issue or the changes).

sergeyklay:
@steffengy done

PR #1137, opened 2015-12-03
Update install

onshorechet:
The command referenced by $? returns 0 on success not 2.  2 is an error code of some sort.

### DIFF
--- a/install
+++ b/install
@@ -8,8 +8,7 @@ if [ ! -f lemon ]; then
 	gcc -w lemon.c -o lemon
 fi
 
-TEST=`re2c -v`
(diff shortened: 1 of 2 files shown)


steffengy:
This breaks all travis builds, so it obviously isn't correct that way.

onshorechet:
So I don't think the exit code can be relied on the way it was.  This commit uses the output from the version, which should be more reliable.  The old version breaks on centos7.

steffengy:
That is much better. Thank you!
I would also have merged the previous version of this fix, if it wasn't breaking more than it's fixing.

PR #1139, opened 2015-12-04
Composer >= 5.4

ovr:


ovr:
@steffengy

sergeyklay:
`master`? Phalcon `2.0.x` no longer will be compiled with new Zephir versions?

steffengy:
@sergeyklay 
I talked with @andresgutierrez about that, and he also was sharing my
opionion that anyone using PHP 5.3 is pretty much reckless (many security issues and no EOL since more than a year).

I guess that either means that phalcon is also dropping PHP5.3 or that it
might be (as I mentioned several times) make sense to use tagged releases
instead of the master branch for phalcon.

hylent:
It's time to support Trait :smiley:

PR #1141, opened 2015-12-05
feat(parser): Initial implementation of parser as PHP ext

steffengy:
Implement zephir's parser as PHP extension supporting 7.0.0>=PHP>5.3.
This removes the overhead of callling an extra executable per file.
## Other Consequences:
- this removes the json-c extension
- on each build the parser can easily be built and is always up to date


Green-Cat:
:+1:

andresgutierrez:
+1

steffengy:
It's because install_zephir.sh contains the part, which compiles & installs the parser, remove that and it should work.

Issue #1142, opened 2015-12-09
Makefile:183: recipe for target 'parser/parser.lo' failed


valVk:
commit c9735cc7ceb1e80908165c3eded57206f5a7319e

``` git
Merge: 9029835 712a1a2
Author: Steffen Butzer <steffengy@users.noreply.github.com>
Date:   Sun Dec 6 00:46:17 2015 +0100

    Merge pull request #1141 from steffengy/parser_ext

    feat(parser): Initial implementation of parser as PHP ext
```

``` bash
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 15.04
Release:        15.04
Codename:       vivid

Linux radxa 3.18.0-rc5+ #2 SMP Fri Nov 28 17:36:47 CST 2014 armv7l armv7l armv7l GNU/Linux
```

All this gives me an error

```
....

Makefile:183: recipe for target 'parser/parser.lo' failed

PHP Warning:  copy(modules/zephir_parser.so): failed to open stream: No such file or directory in /home/rock/cphalcon/zephir/Library/Compiler.php on line 251
Zephir\Exception: The zephir parser extension could not be found or compiled!

at Library/Compiler.php(256)
#0 Library/Compiler.php(269): Zephir\Compiler->compileParser()
#1 Library/Compiler.php(328): Zephir\Compiler->preCompile('test/arithmetic...')
#2 Library/Compiler.php(872): Zephir\Compiler->recursivePreCompile('test')
#3 Library/Compiler.php(1076): Zephir\Compiler->generate(Object(Zephir\Commands\CommandCompile))
#4 Library/Commands/CommandAbstract.php(108): Zephir\Compiler->compile(Object(Zephir\Commands\CommandCompile))
#5 Library/Bootstrap.php(200): Zephir\Commands\CommandAbstract->execute(Object(Zephir\Config), Object(Zephir\Logger))
#6 compiler.php(21): Zephir\Bootstrap::boot()
#7 {main}
```

But phalcon compiled well on exact same configuration

steffengy:
That output doesn't seem complete, can you post the entire output?
Before `Makefile:183: recipe for target 'parser/parser.lo' failed` should be some  more detailed error output.

valVk:
```
rock@radxa:~/cphalcon/zephir$ ./bin/zephir compile
zephir_parser extension not loaded, compiling it
Parser statistics: 127 terminals, 99 nonterminals, 459 rules
[...]

PR #1160, opened 2016-01-04
fix(parser): only include strndup, when required, openSUSE

mruz:
openSUSE MACRO https://github.com/phalcon/zephir/issues/1142#issuecomment-167769011


sergeyklay:
Looks good to me

PR #1161, opened 2016-01-06
Fix fwrite() optimizer in php 7

hylent:
Invalid first param type of zephir_fwrite(), zval \* required, but zval provided.


steffengy:
Thanks

Issue #1121, opened 2015-10-31
[NFR] isProperty shortcut


sergeyklay:
Without isProperty shortcut we usually have code like:

``` zephir
namespace App;

class MyClass
{
    protected fallback = true {
        set
    };

    protected cancelable = true;

    public function isCancelable()
    {
        return this->cancelable;

        // or something like this
        // return (boolean) this->cancelable;
    }

    public function setFallback(boolaen fallback)
    {
        let this->fallback = fallback;
    }

    public function isFallback()
    {
        return this->fallback;
    }
}
```

My proposal is to introduce `is<property>` shortcut. 
So we can write the same code using shortcuts as follows:

``` zephir
namespace App;

class MyClass
{
    protected fallback = true {
        set, is
    };

    protected cancelable = true {
        is
    };
}
```

steffengy:
I personally think that this is not necessary, 
I won't implement it for now, which doesn't mean that this won't be merged.

mruz:
What is the difference to `get` and `getFallback()`?

``` zep
namespace App;

class MyClass
{
    protected fallback = true {
        set, get
    };

}
```

sergeyklay:
``` php
if ($component->isFalback())
```

better than

``` php
if ($component->getFalback())
```

Issue #417, opened 2014-05-26
Setters function doesn't return current object


alkana:
On cphalcon 2.0.0b assets test-unit. (https://github.com/phalcon/cphalcon/blob/2.0.0/unit-tests/AssetsTest.php)

With the latest version I have this error: "Call to a member function setLocal() on a non-object in /home/vagrant/cphalcon-2.x.x/unit-tests/AssetsTest.php on line 155"

It's because, setters doesn't return the current object when you declare functions like that: "protected _local = true { get, set };" (line: 32, file: https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/assets/collection.zep)

complete stack trace:

```
Starting test 'AssetsTest::testAssetsManager'.
PHP Fatal error:  Call to a member function setLocal() on a non-object in /home/vagrant/cphalcon-2.x.x/unit-tests/AssetsTest.php on line 155
PHP Stack trace:
PHP   1. {main}() /usr/bin/phpunit:0
PHP   2. PHPUnit_TextUI_Command::main() /usr/bin/phpunit:46
PHP   3. PHPUnit_TextUI_Command->run() /usr/share/php/PHPUnit/TextUI/Command.php:129
PHP   4. PHPUnit_TextUI_TestRunner->doRun() /usr/share/php/PHPUnit/TextUI/Command.php:176
PHP   5. PHPUnit_Framework_TestSuite->run() /usr/share/php/PHPUnit/TextUI/TestRunner.php:349
PHP   6. PHPUnit_Framework_TestSuite->run() /usr/share/php/PHPUnit/Framework/TestSuite.php:705
PHP   7. PHPUnit_Framework_TestSuite->run() /usr/share/php/PHPUnit/Framework/TestSuite.php:705
PHP   8. PHPUnit_Framework_TestSuite->runTest() /usr/share/php/PHPUnit/Framework/TestSuite.php:745
PHP   9. PHPUnit_Framework_TestCase->run() /usr/share/php/PHPUnit/Framework/TestSuite.php:775
PHP  10. PHPUnit_Framework_TestResult->run() /usr/share/php/PHPUnit/Framework/TestCase.php:783
PHP  11. PHPUnit_Framework_TestCase->runBare() /usr/share/php/PHPUnit/Framework/TestResult.php:648
PHP  12. PHPUnit_Framework_TestCase->runTest() /usr/share/php/PHPUnit/Framework/TestCase.php:838
[...]

PR #1174, opened 2016-01-27
Add magic shortcuts: 'is', 'add'

mruz:
``` zep
class Person {

  protected name { set, get  };
  protected age { set, get };
  protected male = true { set, is };
  protected phone { add, get };
  protected email { add, get };
}
```

``` php
$person = new Person;
$person->setName('Julia');
$person->setAge(21);
$person->setMale(false);

$person
    ->addPhone('0123456789')
    ->addEmail('julia@example.com');

echo 'Name: ' . $person->getName() .
     'Age: ' . $person->getAge() .
     'Sex: ' . ($person->isMale() ? 'Male' : 'Female') .
     'Phone: ' . $person->getPhone() .
     'Email: ' . $person->getEmail();
```


mruz:
@steffengy what do you think about this? Refs https://github.com/phalcon/zephir/issues/417

sergeyklay:
Refs https://github.com/phalcon/zephir/issues/1121

mruz:
@andresgutierrez what do you think about it?

Jurigag:
Why exactly this is still not merged ?

Jurigag:
@andresgutierrez @sergeyklay Well asking above question again i guess.

sergeyklay:
@Jurigag https://github.com/phalcon/zephir/issues/1121#issuecomment-154507884

Jurigag:
Yea but **THIS IS PR**, not issue meaning implemented by @mruz, so why not merge it?

sergeyklay:
Because there are no enough quorum

Jurigag:
Oh, okay.

steffengy:
also the implementation of adding `is` as alias for `get` is quite strange.
`is` should always explicitly return a boolean.

Before merging this we should also come to a consensus
on what "magic shortcuts" we want.

I do not want several hundreds of them, but do not see an issue with adding
those reducing boilerplate.

These uncertainties in terms of the design are the main reason
this is still not merged (and also the reason why this should not be merged
until these are resolved).

mruz:
Originally there was only `add` "magic shortcuts" in this PR.
I asked about difference between `is` and `get` https://github.com/phalcon/zephir/issues/1121#issuecomment-177171607 @sergeyklay just said:
> better than

so I just created an alias to `get`. I didn't know about uncertainties, I'll push a new implementation for `is`.

mruz:
@steffengy do we need to use backend->fetchProperty for returning bool in `is` shortcut? https://github.com/mruz/zephir/blob/e94c202b9a690234d6d4dbdceb919bb066123eb8/Library/Statements/ReturnStatement.php#L73

steffengy:
@mruz depends if the generated code is different for ZE3.
since it currently isn't in any backend-specific class, I guess it's not.

mruz:
@steffengy how about this implementation for the `is`?

sergeyklay:
Cc: @phalcon/zephir-team

flyingangel:
Good idea. 
But "add" does the same as "set" so there is no use
"is" should only return a boolean not a value : return (bool)[prop] === true
@sergeyklay

mruz:
When I created it, `set` did not return anything, so I added `add`, `is` should return bool https://github.com/phalcon/zephir/pull/1174/files#diff-6831a39c4308ea680610729532fc6bb7R43.

mruz:
@sergeyklay I can't rebuild failed build https://travis-ci.org/phalcon/zephir/builds/425215861?utm_source=github_status&utm_medium=notification, could you? or do we still need that PR?

sergeyklay:
@mruz Could you please submit this PR to the 0.10.x branch? Also could you please squash commits into one to keep history clean.

PR #1175, opened 2016-01-27
Allow users to configure their phpize and php-config location from their projects

Izopi4a:
allow users to add those options into their project/config.json file

```
"environment": {
        "phpize": "_PATH_TO_PHPIZE_", #ex /opt/php-5615/bin/phpize
        "php-config": "_PATH_TO_PHP-CONFIG", #ex /opt/php-5615/bin/php-config
        "alwaysCompileParser" : true 
}
```

the reason i added `alwaysCompileParser` is because i saw couple of issues about people trying php 7, then they go back to php5 and in this case compiling the parser throws errors. Enabling this should prevent any kind of errors of this kind.

also removed the `@unlink` stuff since i hate it :D

### DIFF
--- a/Library/Compiler.php
+++ b/Library/Compiler.php
@@ -126,12 +126,10 @@ class Compiler
     public $backend;
 
     /**
-     * Compiler constructor
-     *
+     * Compiler constructor.
      * @param Config $config
      * @param Logger $logger
      * @param BaseBackend $backend
-     * @throws Exception
      */
     public function __construct(Config $config, Logger $logger, BaseBackend $backend)
     {
@@ -168,9 +166,8 @@ class Compiler
 
     /**
      * Adds a function to the function definitions
-     *
      * @param FunctionDefinition $func
-     * @param array $statement
+     * @param null $statement
      * @throws CompilerException
      */
     public function addFunction(FunctionDefinition $func, $statement = null)
@@ -197,6 +194,7 @@ class Compiler
         if ($this->parserCompiled === true) {
             return true;
         }
+
 
         $extFile = null;
         $currentDir = realpath(dirname(__FILE__)). DIRECTORY_SEPARATOR  . '..' .DIRECTORY_SEPARATOR . 'parser' . DIRECTORY_SEPARATOR ;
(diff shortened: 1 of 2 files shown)


mruz:
I think `alwaysCompileParser` is unnecessary due to `parser-compiled`

``` sh
zephir generate --parser-compiled=force
```

Library/Compiler.php
```diff
@@ -168,9 +166,8 @@ protected function checkRequires()
 
     /**
      * Adds a function to the function definitions
-     *
      * @param FunctionDefinition $func
-     * @param array $statement
+     * @param null $statement
```
sergeyklay, line 170: rly?

Izopi4a, line 170: i am used to see what's the default parameter, but if you agree i will change it back to array

Izopi4a:
i saw the `force compile` option, however i think project specific stuff can make the developer live easier. 

Easy to remove if you want.

steffengy:
I do not quite see a point for adding a configuration for phpize/php-config.

The alwaysCompileParser stuff is unnecessary: A proper parser detection
implementation, which currently unfortunately nobody seems to have time to implement,
would just make it unnecessary.

Izopi4a:
the point is that my ubuntu comes with php 5.3 and no updates.
I am always running latest php version, since in php 7 you have to use `zephir build` i am not able to specify my php version. In this way i can specify where the extension should be installed. 

https://github.com/phalcon/cphalcon/issues/11203#issuecomment-173178890

steffengy:
It doesn't make sense to configure the path to PHP in a configuration which is specific to zephir,
as this introduces a dependency to your local system or to a specific directory layout, which isn't what project-specific configuration files are for. That is like trying to use a makefile configured on another system...

If you want to use a custom path to php-config/phpize that's something to be configured by modifying your path (as zephir doesn't support environmental variables for this kind of stuff).
Alternatively you should'be fine to use `zephir generate --backend=ZendEngine3`
and do the compilation step manually, which is probably the worse solution, except
if you need more flexibility anyways.

Izopi4a:
i am not gonna argue, but i am pretty sure people will start crying over this. Many people are compiling their php's. I am pretty sure i am not the only one, now i am force to edit my `/etc/alternatives` directory every single time i want to compile a php7 extension. 

Thank you for the feedback.

steffengy:
> Many people are compiling their php's

Again this isn't a problem at all.
It's more than trivial to setup a short wrapper script, which makes sure everything is setup
for the correct PHP version and then calls zephir build (or passes your arguments to zephir).

It could make sense though to allow users to specify something like `--php_dir=/tmp/php7` like the
PHP build system uses, which wouldn't misuse project settings for system-specific settings.

PR #1177, opened 2016-01-30
Easy access to Zephir function name in AbstractOptimizer

flip111:


sergeyklay:
@flip111 What issue does it solve?

Library/Optimizers/OptimizerAbstract.php
```diff
@@ -21,12 +21,20 @@
...
 use Zephir\Call;
 use Zephir\CompilationContext;
+use Zephir\Utils;
 
 /**
  * Class OptimizerAbstract
  */
 abstract class OptimizerAbstract
 {
+	protected $functionName;
```
sergeyklay, line 31: Spaces must be used to indent lines; tabs are not allowed

Library/Optimizers/OptimizerAbstract.php
```diff
@@ -21,12 +21,20 @@
...
+
+	public function __construct() {
```
sergeyklay, line 33: Opening brace should be on a new line

sergeyklay (changes requested):

flip111:
I forgot the specific case, but in case you have several optimizers calling the same function you can now pass information to C which function it was.

Issue #1179, opened 2016-02-04
Getter/Setter shortcuts - Zephir compiler only capitalizing, not camelizing property name to create PHP getter/setter methods


tabalchi:
I have several DB table column names with underscores (e.g. first_name, last_name, etc.). When writing the model code in PHP, it is normal to write getters/setters using camelization of the column names. However, when I rewrote the model code in Zephir and used the getter/setter shortcuts, the rest of my PHP code that uses the model started crashing. That's when I discovered that Zephir is simply _capitalizing_ the property names to derive the getter/setter method names when shortcuts are used _instead of camelizing_.

For example, if the property name is `first_name` (must match column name for model to work), the getter ends up being called `getFirst_name()` instead of `getFirstName()` which is more conventional. Same with the setter.

The fix is quite simple. In the file

`zephir/Library/CompilerFile.php` do the following:

In function

``` php
_processShorcuts()
```

Inside the switch-case statements for **get** and **set**,

``` php
            switch ($shortcut['name']) {
                case 'get':
                    ....

                case 'set':
                    ....
```

Replace the line 

``` php
'get' . ucfirst($name),
```

with 

``` php
'get' . Utils::camelize($name),
```

and make the same change for 'set'

Below is the code example.
In writing the model code in Zephir, I wrote the following:

``` php
class Customer extends \Phalcon\Mvc\Model
{
    protected first_name {
        set, get
    };
}

```

The corresponding .c file has the following in it:

``` php
ZEPHIR_INIT_CLASS(MyProject_Models_Customer) {

    ZEPHIR_REGISTER_CLASS_EX(MyProject\\Models, Customer, myproject, models_customer, zephir_get_internal_ce(SS("phalcon\\mvc\\model") TSRMLS_CC), myproject_models_customer_method_entry, 0);

    zend_declare_property_null(myproject_models_customer_ce, SL("first_name"), ZEND_ACC_PROTECTED TSRMLS_CC);

    return SUCCESS;
}

PHP_METHOD(MyProject_Models_Customer, setFirst_name) {
    zval *first_name;
    zephir_fetch_params(0, 1, 0, &first_name);
    zephir_update_property_this(this_ptr, SL("first_name"), first_name TSRMLS_CC);
}

PHP_METHOD(MyProject_Models_Customer, getFirst_name) {
    RETURN_MEMBER(this_ptr, "first_name");
}
```

As can be seen, the getter/setter methods created are **setFirst_name** and **getFirst_name**.

[...]

PR #1183, opened 2016-02-08
Pull request for issue #1179 on phalcon/zephir: Use camelization for getter/setter shortcut function names

tabalchi:
This change results in auto-generation of getter/setter function names using **camelization** of property name instead of first letter capitalization when processing getter/setter shortcut directives. Example: If property is

``` php
protected my_sample_property {
    get, set
};
```

then the auto-generated PHP setter/getter function names will now be generated as

``` php
getMySampleProperty()
setMySampleProperty()
```

which is more conventional, rather than

``` php
getMy_sample_property()
setMy_sample_property()
```

### DIFF
--- a/Library/CompilerFile.php
+++ b/Library/CompilerFile.php
@@ -341,7 +341,7 @@ class CompilerFile
                     $classDefinition->addMethod(new ClassMethod(
                         $classDefinition,
                         array('public'),
-                        'get' . ucfirst($name),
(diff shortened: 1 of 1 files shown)


sergeyklay:
Looks good to me

Green-Cat:
:+1:

tabalchi:
Thanks!

PR #1189, opened 2016-02-13
[Refactoring] getVariableCodePointer, getVariableCode must be defined…

ovr:
… in BaseBackend

### DIFF
--- a/Library/Backends/ZendEngine2/Backend.php
+++ b/Library/Backends/ZendEngine2/Backend.php
@@ -25,7 +25,7 @@ class Backend extends BaseBackend
     }
 
     /**
-     * Get a double pointer to the variable
+     * {@inheritdoc}
      */
     public function getVariableCodePointer(Variable $variable)
     {
@@ -35,6 +35,9 @@ class Backend extends BaseBackend
         return '&' . $variable->getName();
     }
 
+    /**
+     * {@inheritdoc}
+     */
     public function getVariableCode(Variable $variable)
     {
         if ($variable->isLocalOnly() && !in_array($variable->getType(), array('int'))) {
--- a/Library/Backends/ZendEngine3/Backend.php
+++ b/Library/Backends/ZendEngine3/Backend.php
@@ -23,6 +23,9 @@ class Backend extends BackendZendEngine2
         return true;
     }
 
+    /**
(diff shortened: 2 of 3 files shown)


ovr:
ping @steffengy

steffengy:
I'm still not quite sure if that's a good idea.

Theoretically a backend wouldn't necessarily need to define these methods,
since that's specific to zend or rather C backends, which might also
depend further on the implementation.

For now it seems appropriate though, since additional backend support seems very unlikely.

ovr:
For now Zephir is using this methods from Backed
For most cases `$backend` property defined as `BaseBackend`
And this methods is needed...

steffengy:
For the current case I agree with you, the code base is anyways too close to Zend,
so it is anyways a wasted effort to try to separate that, while leaving the other areas
untouched.

Library/Backends/ZendEngine2/Backend.php
```diff
@@ -35,6 +35,9 @@ public function getVariableCodePointer(Variable $variable)
         return '&' . $variable->getName();
     }
 
+    /**
+     * {@inheritdoc}
```
sergeyklay, line 39: > Currently some applications have DocBlocks containing just the {@inheritDoc} inline tag to indicate that their complete contents should be inherited. This usage breaks with the PHPDoc Standard as summaries cannot contain inline tags and inheritance is automatic; you do not need to define a special tag for it.
> 
> However, it does make clear that an element has been explicitly documented (and thus not forgotten). As such we are working to include a new (normal) tag in the PHPDoc Standard @inheritDoc that will serve that purpose.

http://phpdoc.org/docs/latest/guides/inheritance.html

ovr, line 39: @sergeyklay 

It's notice that it can breaks some applications but for now it's working and most of all projects are using it
For example I am using inline plugin for phpstorm and atom for it

image

image

Review sf2
https://github.com/symfony/symfony/search?utf8=%E2%9C%93&q=%7B%40inheritDoc%7D

image

or Zf2 component
https://github.com/zendframework/zend-paginator/search?utf8=%E2%9C%93&q=%7B%40inheritDoc%7D

image

PR #1196, opened 2016-03-03
Add "MongoDB\Driver\manage" for mongodb

danclive:
Add "MongoDB\Driver\manage" for mongodb.
see: http://php.net/manual/en/set.mongodb.php
http://pecl.php.net/package/mongodb
https://github.com/mongodb/mongo-php-driver


danclive:
@steffengy OK, I have added a line end of the file.

Issue #1198, opened 2016-03-05
Use of undefined constant PHP_BINARY


kylekatarnls:
Hi, I get this error on Travis CI when I try to test a PHP extension made with Zephir:

```
PHP Notice:  Use of undefined constant PHP_BINARY - assumed 'PHP_BINARY' in /home/travis/build/kylekatarnls/csrfprotect/.build-travis/zephir/Library/Compiler.php on line 296
```

As PHP_BINARY is not set in every environment, what if we just test if PHP_BINARY exists and else use "php"?

steffengy:
PHP_BINARY is available since PHP5.4.
If you get that error you're using an older version, which is not supported by zephir.

kylekatarnls:
It seems not on Travis, this is a PHP 5.4 test:
https://travis-ci.org/kylekatarnls/csrfprotect/jobs/113890846

steffengy:
then travis is probably not using PHP5.4 there.
https://3v4l.org/h4RDb

**edit issue found**:
you install another PHP via `apt get`, which isn't very clever in that situation:

```
Get:7 http://us-central1.gce.archive.ubuntu.com/ubuntu/ precise-updates/main php5-common amd64 5.3.10-1ubuntu3.21 [1,776 kB]
```

kylekatarnls:
Oh thanks, I tried many things and forget to remove that. I try to cleanup a bit the travis script.

PR #1199, opened 2016-03-05
#1198 Use php if PHP_BINARY is not defined

kylekatarnls:


steffengy:
PHP_BINARY is available in all recent PHP versions (since 2014)

PR #1201, opened 2016-03-06
Now, the developer can load the custom prototypes.

danclive:
Recently, I in the development of a framework, using the new mongodb driver. The driver uses the namespace, file name cannot extend to name, on the "zephir/prototypes" doesn't work very well. Therefore I think of a better way to provide an option to let me decide which directory is loaded. By my own and maintain these, but not all in the zephir/prototypes".  
In the configuration file:

``` json
{
    "prototype-dir": {
        "mongodb": "prototypes/mongodb"
    }
}
```

If "mongodb extension" was not load, the file in "prototypes/mongodb" will be loaded.

### DIFF
--- a/Library/Compiler.php
+++ b/Library/Compiler.php
@@ -973,6 +973,29 @@ class Compiler
                 }
             }
 
+            /**
+             * Load customer additional extension prototypes
+             */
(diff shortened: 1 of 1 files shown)


steffengy:
that seems fine & like a very good idea.
Could you please squash your commits though?

PR #1215, opened 2016-03-19
Update Compile script for Windows x64 environment

kabadna:


steffengy:
thanks, I guess this is now relevant since PHP7 supports x64 on Win.

PR #1227, opened 2016-04-01
Delete symbol_table usage for zend_fcall_info strut (php 7.1.0-dev).

hervems:
See commit in php 7 source / master of Dmitry Stogov dmitry@zend.com:
"PHP-7 doesn't support symbol_table substitution for functions".
Today it's an assert into zend_execute_API.c: line 855 into php 7.0.5 source.

### DIFF
--- a/kernels/ZendEngine3/fcall.c
+++ b/kernels/ZendEngine3/fcall.c
@@ -420,7 +420,6 @@ int zephir_call_user_function(zval *object_pp, zend_class_entry *obj_ce, zephir_
 	fci.params         = NULL;
 	//fci.params: Passed as separate parameter to prevent the need to convert zval ** to zval *
 	fci.no_separation  = 1;
-	fci.symbol_table   = NULL;
 
 	fcic.initialized = 0;
 	fcic.function_handler = NULL;
--- a/kernels/ZendEngine3/fcall.c
+++ b/kernels/ZendEngine3/fcall.c
@@ -420,6 +420,9 @@ int zephir_call_user_function(zval *object_pp, zend_class_entry *obj_ce, zephir_
 	fci.params         = NULL;
 	//fci.params: Passed as separate parameter to prevent the need to convert zval ** to zval *
 	fci.no_separation  = 1;
+#if PHP_VERSION_ID < 70100
+	fci.symbol_table   = NULL;
+#endif
 
 	fcic.initialized = 0;
 	fcic.function_handler = NULL;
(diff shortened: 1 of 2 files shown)


kernels/ZendEngine3/fcall.c
```diff
@@ -420,7 +420,6 @@ int zephir_call_user_function(zval *object_pp, zend_class_entry *obj_ce, zephir_
 	fci.params         = NULL;
 	//fci.params: Passed as separate parameter to prevent the need to convert zval ** to zval *
 	fci.no_separation  = 1;
-	fci.symbol_table   = NULL;
```
steffengy, line 423: is there any reason you removed this line?

hervems, line 423: Because, in PHP 7.0.X, it's deprecated

```
/* PHP-7 doesn't support symbol_table substitution for functions */
ZEND_ASSERT(fci->symbol_table == NULL);
```

and in PHP 7.1.X (see master of PHP, it's deleted), see :

zend_API.h :

typedef struct _zend_fcall_info {
    size_t size;
    zval function_name;
    zval *retval;
    zval *params;
    zend_object *object;
    zend_bool no_separation;
    uint32_t param_count;
} zend_fcall_info;

so, it's not necessary to use it.

I saw in Zephir, no impact to delete this line and to set call->symbol_table
= NULL in "
extended/fcall.c".

2016-05-05 12:43 GMT+02:00 Steffen Butzer notifications@github.com:

steffengy, line 423: Just because you saw no impact it doesn't mean there won't be one.
Removing that would be ok if there was no 7.X version using it.

Even your own code example shows why it's not a good idea:
`ZEND_ASSERT(fci->symbol_table == NULL);` might fail since `->symbol_table`
might be initialized to a not-null value (due to not-zeroed memory allocation).

So we'll have to wrap this into ifdef's for 7.1.X. 
(since that won't land in PHP7.0.X exactly for that backwards-compatibility argument above)

hervems, line 423: If you follow the execution path after the line :

fci.symbol_table = null in zephir_call_user_function

the unique usage of fci->symbol_table is in extended/fcall.c, line 144.

we never go into the code of PHP in this context, so why initialize an
attribut when it's not used.

So I'm agree I can add a :

#if PHP_VERSION_ID < 70100
  ...
#endif

to be more secure.

2016-05-07 1:56 GMT+02:00 Steffen Butzer notifications@github.com:

steffengy, line 423: In that case that might be true, but generally you should always prefer the safe approach insteadof the one where you might miss additional dependencies.

steffengy:
thanks

PR #1231, opened 2016-04-07
Force compile parser fix

crocodile2u:
Fixes issue with spawning PHP processes in a infinite loop when --parser-compiled=force is specified; also fixes parser recompilation being skipped even when --parser-compiled=force is specified.


crocodile2u:
Anybody interested in this? It's quite an error, to be honest, and this fix made my life easier.

sergeyklay:
@steffengy

steffengy:
Since this seems like the cleanest way to fix it (even though there're still problems with parser-compilation), i'll merge this. We might still have to improve i later on.

Issue #1202, opened 2016-03-07
Error in _zephir_call_user_function in ZendEngine3 fcall.c


SerafimArts:
Environment:
1) php 7.0 x86 nts
1.1) Trying with TS too
2) VS 2015 (version 14.0)
3) Zephir v0.8
3.1) Trying with master branch too
4) Windows 7 x64

Code:

```
namespace App;

class Version
{
    public static function get()
    {
        return "0.0.1 super beta";
    }
}
```

Then run `zephir compile`:

```
...>zephir compile
Copying new kernel files...
Preparing for PHP compilation...
Preparing configuration file...
Compiling...
```

No errors but has compile.log:

```
"Cl.exe" / FI main blah blah blah...
...
object.c
operators.c
require.c
string.c
D:\Projects\Morph\bin\ext\kernel\object.c (65): warning C4090: dgЄzhЁp: a §lҐ Єў "ЁdЁЄ v®al" const "
time.c
variables.c
"Cl.exe" /... blah blah blah ...

Microsoft (R) Windows (R) Resource Compiler Version 10.0.10011.16384
Copyright (C) Microsoft Corporation. All rights reserved.

D:\Projects\Morph\bin\ext\Release\array.obj: warning LNK4042: Set the object more than once; excess are not counted
D:\Projects\Morph\bin\ext\Release\fcall.obj: warning LNK4042: Set the object more than once; excess are not counted
   Create a library D:\Projects\Morph\bin\ext\Release\php_app.lib and object D:\Projects\Morph\bin\ext\Release\php_app.exp
fcall.obj: error LNK2019: unresolved external reference to the symbol in _zephir_call_function_opt _zephir_call_user_function function
D:\Projects\Morph\bin\ext\Release\php_app.dll: fatal error LNK1120: unresolved external elements: 1
```

And only php_app.dll.manifest and php_app.dll.res in Release dir, no binaries
='(

SerafimArts:
I trying to build extension over **clear** virtual machine:
1) Windows 10 x64 Enterprise 2015 LTSB
2) Zephir release snapshot v0.8
3) PHP 7.0.4 Non Thread Safe x86
4) PHP 7 (x86) develpack from http://windows.php.net/downloads/releases and php 7 sdk from http://windows.php.net/downloads/php-sdk/ (with binary tools copied inside sdk path)
5) VisualStudio 2015 Express with C++ support and Extensibility tools update 1

And i have this error again:
[...]

PR #1234, opened 2016-04-12
Windows PHPIZE path issues resolved

umermalik129:


steffengy:
@umermalik129 doesn't seem to work, tests (appveyor) are failing.

umermalik129:
@steffengy where code is breaking? I will try to fix it

steffengy:
https://ci.appveyor.com/project/phalcon/zephir/build/1.0.628

umermalik129:
@steffengy what is this fatal error? "Fatal error: Class 'Test\Arithmetic' not found in C:\projects\zephir\unit-tests\Extension\ArithmeticTest.php on line 28"

steffengy:
I think you should probably take a look at the context (it just means it couldnt load the DLL - so the class does not exist - in that case because compilation fails)

```
C:\projects\zephir\ext\test\testinterface.zep.c : fatal error C1083: Cannot open compiler generated file: 'Release_TS\projects\zephir\ext\test\testinterface.zep.obj': No such file or directory
```

which hints that compilation fails ^^

umermalik129:
To test the code after changing I have to create PR or there is other way? I don't want to setup own CI server

steffengy:
@umermalik129 
you could try to compile the zephir test extension yourself (appveyor.yml should hint at that)
any commit you add to the PR (push to your winfix branch) also gets automatically tested by the CI.

umermalik129:
@steffengy I reverted `processAddSources` related changes and now the build is successful 

I think we only need to change the paths that will go inside release directory

umermalik129:
@steffengy will it work?

steffengy:
yeah sure tests work, but static builds are broken, just as before, without patching the configure script.

umermalik129:
You mean Travis CI? appveyor builds successfully and I think it is related to windows build

steffengy:
No I mean by reverting the changes you did you effectively get appveyor to work, but
break in case of #1202.
Anyways the current state in master works in all cases I'm aware of,
and you really shouldn't waste your time in fixing something that isn't broken,
atleast if not being done systematically.

PR #1240, opened 2016-04-18
Added optional delimeter to camelize/uncamelize

sergeyklay:
This PR aims to solve issues like this https://github.com/phalcon/cphalcon/issues/10396

### DIFF
--- a/Library/Optimizers/FunctionCall/UncamelizeOptimizer.php
+++ b/Library/Optimizers/FunctionCall/UncamelizeOptimizer.php
@@ -45,8 +45,15 @@ class UncamelizeOptimizer extends OptimizerAbstract
             return false;
         }
 
-        if (count($expression['parameters']) != 1) {
-            throw new CompilerException("'uncamelize' only accepts one parameter");
+        if (count($expression['parameters']) < 1 || count($expression['parameters']) > 2) {
+            throw new CompilerException("'uncamelize' only accepts one or two parameters");
+        }
+
+        $delimiter = 'NULL ';
+        if (count($expression['parameters']) == 2) {
+            if ($expression['parameters'][1]['parameter']['type'] == 'null') {
+                unset($expression['parameters'][1]);
+            }
         }
 
         /**
@@ -64,12 +71,17 @@ class UncamelizeOptimizer extends OptimizerAbstract
         $symbolVariable->setDynamicTypes('string');
 
         $resolvedParams = $call->getReadOnlyResolvedParams($expression['parameters'], $context, $expression);
+        
+        if (isset($resolvedParams[1])) {
+            $delimiter = $resolvedParams[1];
+        }
+
         if ($call->mustInitSymbolVariable()) {
             $symbolVariable->initVariant($context);
         }
 
         $symbol = $context->backend->getVariableCode($symbolVariable);
-        $context->codePrinter->output('zephir_uncamelize(' . $symbol . ', ' . $resolvedParams[0] . ');');
+        $context->codePrinter->output('zephir_uncamelize(' . $symbol . ', ' . $resolvedParams[0] . ', ' . $delimiter . ' );');
 
         return new CompiledExpression('variable', $symbolVariable->getRealName(), $expression);
     }
--- a/kernels/ZendEngine2/string.c
+++ b/kernels/ZendEngine2/string.c
@@ -252,16 +252,25 @@ void zephir_fast_join_str(zval *return_value, char *glue, unsigned int glue_leng
 }
 
 /**
- * Convert dash/underscored texts returning camelized
+ * Convert dash/underscored texts returning camelized (an optional delimiter can be specified)
  */
-void zephir_camelize(zval *return_value, const zval *str) {
+void zephir_camelize(zval *return_value, const zval *str, const zval *delimiter) {
 
 	int i, len, first = 0;
 	smart_str camelize_str = {0};
-	char *marker, ch;
+	char *marker, ch, delim;
 
 	if (unlikely(Z_TYPE_P(str) != IS_STRING)) {
 		zend_error(E_WARNING, "Invalid arguments supplied for camelize()");
+		RETURN_EMPTY_STRING();
+	}
+
+	if (delimiter == NULL || Z_TYPE_P(delimiter) == IS_NULL) {
+		delim = '_';
+	} else if (Z_TYPE_P(delimiter) == IS_STRING && Z_STRLEN_P(delimiter) == 1) {
+		delim = *(Z_STRVAL_P(delimiter));
+	} else {
+		zend_error(E_WARNING, "Second argument passed to the camelize() must be a string of one character");
 		RETURN_EMPTY_STRING();
 	}
 
@@ -274,7 +283,7 @@ void zephir_camelize(zval *return_value, const zval *str) {
 
 		if (first == 0) {
 
-			if (ch == '-' || ch == '_') {
+			if (ch == delim) {
 				continue;
 			}
 
@@ -283,7 +292,7 @@ void zephir_camelize(zval *return_value, const zval *str) {
 			continue;
 		}
 
-		if (ch == '-' || ch == '_') {
+		if (ch == delim) {
 			if (i != (len - 1)) {
 				i++;
 				ch = marker[i];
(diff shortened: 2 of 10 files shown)


Library/Optimizers/FunctionCall/UncamelizeOptimizer.php
```diff
@@ -64,12 +71,17 @@ public function optimize(array $expression, Call $call, CompilationContext $cont
         $symbolVariable->setDynamicTypes('string');
 
         $resolvedParams = $call->getReadOnlyResolvedParams($expression['parameters'], $context, $expression);
+        
+        if (isset($resolvedParams[1])) {
+            $delimer = $resolvedParams[1];
```
virgofx: Maybe spell as "delimeter" as well on line 84.

sergeyklay: Thanks. Fixed

virgofx:
Looks good, thanks :+1:

SidRoberts:
"delim**e**ter" should be "delim**i**ter".

unit-tests/Extension/BuiltIn/StringMethodTest.php
```diff
@@ -23,6 +23,105 @@
 
 class StringMethodTest extends \PHPUnit_Framework_TestCase
 {
+    public function testCamelize()
```
fezfez: You should use data provider (https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers)

sergeyklay: Thanks. Improved

steffengy:
@sergeyklay 
It seems like the function generally would not work with delimiters with a length>1 (since there're many char*-deref compares etc. (won't work with strings of an arbitrary size imo)

sergeyklay:
It need to work with strings of an arbitrary size?

steffengy:
@sergeyklay
Normally not necessarily (even though that could make sense in rare cases of test__test2)
but since you do `delim = "_\0";` that pretty much implies that it should work with arbitrary size
doesn't it? (since you declare it as a string (char*) and later only check one char that seems inconsistent asfar as I noticed when skimming the commit)

sergeyklay:
@steffengy 
Ah :) I understand. No, I would like to add ability to use custom delimiter. One char, no more. So I need to refactor this.

sergeyklay:
@steffengy Could you please explain why now tests fails? I've tested this PR on my PC (PHP 5.4, 5.5, 5.6)  - all ok

steffengy:
@sergeyklay likely related to @andresgutierrez currently fixing some parser stuff
(at the moment still broken), but it's something that's long overdue :)
I'll guess we'll have to wait a bit until that's sorted out, no hurry :)

sergeyklay:
@steffengy 
Nevertheless could you take a look at the current version of PR?

kernels/ZendEngine2/string.c
```diff
@@ -305,19 +311,25 @@ void zephir_camelize(zval *return_value, const zval *str) {
...
  */
-void zephir_uncamelize(zval *return_value, const zval *str) {
+void zephir_uncamelize(zval *return_value, const zval *str, const zval *delimiter) {
 
 	unsigned int i;
 	smart_str uncamelize_str = {0};
-	char *marker, ch;
+	char *marker, ch, delim;
 
 	if (Z_TYPE_P(str) != IS_STRING) {
```
steffengy, line 325: looks good so far, I think it could make sense to add `Z_STRLEN_P(str)>1` here aswell.
(to catch invalid input)

I would recommend holding back this commit until master is fixed again,
then you can use that minor change to trigger the CI builds ^^
(which very likely will pass and make this mergeable)

sergeyklay, line 325: do you mean throw warning on `Z_STRLEN_P(str) <= 1` ?

steffengy, line 325: I'd just do `if (Z_TYPE_P(str) != IS_STRING || Z_STRLEN_P(str) > 1) {`

sergeyklay, line 325: I think you mean something else or I've misunderstood :) Because it's pointless

``` c
    if (Z_TYPE_P(str) != IS_STRING || Z_STRLEN_P(str) > 1) {
        zend_error(E_WARNING, "Invalid arguments supplied for uncamelize()");
        return;
    }
```

sergeyklay, line 325: Probably you meant?

``` c
if (Z_TYPE_P(str) != IS_STRING || Z_STRLEN_P(str) < 1) {
```

steffengy, line 325: @sergeyklay 
If you want to be exact it'd be 
`if (Z_TYPE_P(str) != IS_STRING || Z_STRLEN_P(str) != 1) {`
since only string of length 1 are supported.

sergeyklay, line 325: @steffengy `str` is string that must be camelized/uncamelized. So length of `str` shouldn't be in particular size. At least `str` length might be greater than 1

steffengy, line 325: @sergeyklay 
oops, ofcourse you're right.
I actually meant 
`if (Z_TYPE_P(delimiter) != IS_STRING || Z_STRLEN_P(delimiter) != 1) { <ERROR_MESSAGE> }`

sergeyklay, line 325: @steffengy What do you think about `delim = *(Z_STRVAL_P(delimiter));` in your opinion it is OK?

steffengy, line 325: @sergeyklay 
yeah that part is fine, I would just like some error checking before as in my last comment,
since when someone would be to lazy to read the source code and passes something
unexpected the function will not do what he wants it to do, without him figuring out,
likely leading to unnecessary issues and wasted time on his part.

sergeyklay:
@steffengy I would like to ask you how about the fact that both functions return the same result in case of error.

Now `zephir_uncamelize` returns `void` but `zephir_camelize` returns `RETURN_EMPTY_STRING()`.

steffengy:
@sergeyklay 
RETURN_EMPTY_STRING actually only modifies `return_value` 
(the function-argument passed as zval *)
to point to an empty string.

sergeyklay:
Well I can to use `RETURN_EMPTY_STRING()` in both cases?

steffengy:
Since you have 

```
void zephir_camelize(zval *return_value, const zval *str, const zval *delimiter) {
void zephir_uncamelize(zval *return_value, const zval *str, const zval *delimiter) {
```

which are identical signature wise, sure you can.

sergeyklay:
@steffengy Okay. I've took into account your advice and now the code looks really logical. Unfortunately the tests still failed. Frankly I'm not sure what exactly this test fails, but on my local machine everything works fine, including this test. 

Could you please take another look at the result?  :)

steffengy:
multiple reasons:
- parser still seems broken (for the PHP7 build)
- PSR-2 failures (take a look at any travis build)
- uncamelize during the route test seems to fail (segfault)

sergeyklay:
@steffengy gotcha :)  It seems to me test fails here: https://github.com/phalcon/zephir/blob/master/test/router/route.zep#L327

``` zephir
// Always pass the controller to lowercase
let routePaths["controller"] = uncamelize(realClassName);
```

note `uncamelize(realClassName)` instead of `realClassName->uncamelize()`.

I examined the generated C code and found that in this place is generated:

``` c
zephir_uncamelize(_1$$10, realClassName);
```

instead of:

``` c
zephir_uncamelize(_1$$10, realClassName, NULL);
```

See my changes in `Library/Optimizers/FunctionCall/UncamelizeOptimizer.php`.

I have no idea why this is happening. May be you have any idea?

P.S. Also I've updated runtime (see latest commit).

steffengy:
I wouldn't do any changes on runtime since it's anyways massively outdated and might be broken entirley that way.

Have you done a fullclean? (`zephir_uncamelize(_1$$10, realClassName);` would indicate that you haven't?)

The segfault likely btw is:

```
if (Z_TYPE_P(delimiter) == IS_NULL) {
```

should be

```
if (delimiter == NULL || Z_TYPE_P(delimiter) == IS_NULL) {
```

sergeyklay:
@steffengy 
1. Should I rollback the changes from runtime? 
2. Yes, I did `fullclean`. Travis didn't?

https://github.com/phalcon/zephir/blob/master/generated/ZendEngine_2/router/route.zep.c#L485
https://github.com/phalcon/zephir/blob/master/.travis.yml#L33

steffengy:
@sergeyklay 
1. Travis doesn't need a fullclean, since it always creates a new ext-folder (the ext folder is on gitignore)
2. Reverting that makes sense, since it's unmaintained for 10 months.
3. Take a look at that null-pointer dereferences mentioned above, that's likely the issue why travis fails.

sergeyklay:
@steffengy Well, tests (PHP 5.4 - 5.6) passes. Build fails on ZE3 but this is parser issue

andresgutierrez:
Camelize tests are failing on php7: https://travis-ci.org/phalcon/zephir/jobs/125447795

andresgutierrez:
Following warnings maybe have to be checked:

```
/Users/gutierrezandresfelipe/zafiro/ext/kernel/string.c:586:9: warning: incompatible integer to pointer conversion assigning to 'char *' from 'int' [-Wint-conversion]
                delim = '_';
                      ^ ~~~
/Users/gutierrezandresfelipe/zafiro/ext/kernel/string.c:588:9: warning: incompatible integer to pointer conversion assigning to 'char *' from 'char'; remove * [-Wint-conversion]
                delim = *(Z_STRVAL_P(delimiter));
                      ^ ~~~~~~~~~~~~~~~~~~~~~~~~
/Users/gutierrezandresfelipe/zafiro/ext/kernel/string.c:603:11: warning: comparison between pointer and integer ('int' and 'char *')
                        if (ch == delim) {
                            ~~ ^  ~~~~~
/Users/gutierrezandresfelipe/zafiro/ext/kernel/string.c:612:10: warning: comparison between pointer and integer ('int' and 'char *')
                if (ch == delim) {
                    ~~ ^  ~~~~~
```

sergeyklay:
@andresgutierrez @steffengy Fixed here https://github.com/phalcon/zephir/pull/1246

PR #1248, opened 2016-04-26
New specialized parser

andresgutierrez:
This is a refactor to the existing parser extension in Zephir. It takes advantage of the preferred memory management for php5 (heap allocation) and php7 (stack allocation) for zvals. The approach allows us to reduce the boilerplate and the complexity of supporting both Zend Engines in a single parser. The drawback is that for now both parsers must be maintained. 

Compiling Phalcon on Linux 64bits/php 7.0.5 **before** this refactor requires: 318MB of memory
Compiling Phalcon on Linux 64bits/php 7.0.5 **after** this refactor now requires: 153MB of memory

While 153MB is still high, this refactor reduces approximately 45%-55% memory overall usage when compiling a large Zephir project.

Also, my conclusion is that I must follow this approach in Phalcon in PHQL/Volt parsers.


steffengy:
@andresgutierrez 
On the long term, we should aim for PHP_VERSION_ID or similar
 insteadof php5/php7 since you'll also need a recompilation of the parser
if you update PHP (or it get's updated)

andresgutierrez:
Yes, that's the approach I will pursue on PHQL, if everything works fine I'll port it back to Zephir

Jurigag:
I don't know why all of you are writeing size od **debugged** version of extensions ? You shouldn't never use debugged php and phalcon on production enviroment and for me it's obvious. Non debug version is 4.6 mb.

andresgutierrez:
This is not the size of the executable but the peak memory used in RAM to generate the code.

Jurigag:
Oh okay. Then it's great idea to refractor this, beacause it can be tough sometimes to compile phalcon on vps with 512 mb ram and without ability to create swap.

mruz:
Now when I'm trying to build phalcon on php5 I'm receiving:

``` log
sh build_linux.sh
[   52s] parser..lemon: Can't open this file for reading.
[   52s] error: couldn't generate parser
[   52s] error: Bad exit status from /var/tmp/rpm-tmp.YgjNB6 (%build)
```

https://build.opensuse.org/package/live_build_log/home:mruz/php5-phalcon/openSUSE_13.2/x86_64

andresgutierrez:
@mruz If you are running build_linux.sh manually now you have to pass the parser you want to build: `build_linux.sh php5` and `build_linux.sh php7`

mruz:
Ok, thanks. I'll try.

----- Reply message -----
Od: "Andy Gutierrez" notifications@github.com
Do: "phalcon/zephir" zephir@noreply.github.com
DW: "Mariusz Łączak" mruz@poczta.onet.pl, "Mention" mention@noreply.github.com
Temat: [phalcon/zephir] New specialized parser (#1248)
Data: czw., kwi 28, 2016 17:08

@mruz If you are running build_linux.sh manually now you have to pass the parser you want to build: build_linux.sh php5 and build_linux.sh php7

&mdash;You are receiving this because you were mentioned.Reply to this email directly or view it on GitHub

Jurigag:
We need to do it where ? Just zephir root/parser/parser ? And then just zephir build as always ?

Jurigag:
I guess we have to tell phalcon to use php7 files the same way ? But how exactly ?

Well as i see is just working as it is.

mruz:
Wojtek, only when you rebulid the **parser** manually, eg.

``` sh
cd zephir/parser/parser
sh build_linux.sh php5
cd ..
phpize
./configure --enable-zephir_parser
make
```

Jurigag:
Well already tested it and if i delete parser.php5 or parser.php7 from phalcon it causing error so i guess is just working.

But there are still seg fault on phql parser, hope it will be fixed soon.

Issue #1252, opened 2016-05-09
Change to zephir_camelize omits delimiter


mdular:
[This](https://github.com/phalcon/zephir/commit/182e3137fbf243501e4264417b1dd84641635122#diff-06d533289d695a1ece14625a036899fbL594) change breaks backwards compatibility by omitting a delimiter. The change is about adding a parameter + the docbloc still mentions both delimiters, which is why I believe the change in behaviour was not intentional.

Since cphalcon already has this change tagged, and it breaks projects relying on camelize in phalcon 2.0.11 it would be great to find a speedy solution (fast-track, or alternatively pin cphalcon on a previous version of zephir)

Highlighting ZE2 counterpart for convenience: https://github.com/phalcon/zephir/commit/182e3137fbf243501e4264417b1dd84641635122#diff-2ad045f7146412da38eb613636a47f8fL277

steffengy:
@sergeyklay

Issue #1090, opened 2015-09-20
Incorrect str->camelize()


sergeyklay:
How to reproduce

``` zephir
string str1 = "some_string";
string str2 = "someString";

echo str1->camelize(), PHP_EOL;
echo str2->camelize(), PHP_EOL;
```

**Actual**

```
SomeString
Somestring
```

**Expected**

```
SomeString
SomeString
```

PHP's workaround

``` php
$string1  = "some_string";
$string2  = "someString";

$camelize = function ($string) {
    return implode('', array_map('ucfirst', explode('_', $string)));
};
echo $camelize($string1), PHP_EOL;
echo $camelize($string2), PHP_EOL;
```

```
SomeString
SomeString
```

GavinMcL:
Hi looks good I'm seeing correct camel case in the getters/setters now for camel case table field names.

I've just noticed the classname (and therefore php file name), is also being created in title case when the table name is camel case - e.g for a table profileAddress, I get Profileaddress.php with a class Profileaddress - can this be fixed in the same issue as I guess it must use similar camelize function?

sergeyklay:
@GavinMcL Create please separated issue at Devtools repo

GavinMcL:
ok thanks

PR #1253, opened 2016-05-09
Fixed backwards compatibility for zephir_camelize

sergeyklay:
refs #1252, https://github.com/phalcon/cphalcon/issues/11767


sergeyklay:

mdular:
+2 for speed :)

steffengy:
@sergeyklay 
tests don't seem to pass.
I somehow do not quite like it in terms of coding style, but I think it cannot really
be improved in terms of easily finding a nicer way to "write it down".

sergeyklay:
@steffengy Sorry, it is my typo (missed semicolon). I've fixed the PR.

Yes, I totally agree with you about the coding style, but I think it's the most painless way to restore the old behavior.

steffengy:
@sergeyklay 
What do you think of turning the delimiter parameter to a character mask as in `ltrim`?
This essentially would be using a `char *` with the default value of `-_`.
I think that'd be the cleanest solution and actually not hard to refactor.

If you don't want to do it that way yet, you can merge this to fix the upstream breakage
and please file a new issue :)

sergeyklay:
@steffengy Ok. I hope now is more appropriate. Could you please review?

steffengy:
@sergeyklay 
Yeah thanks for that, I'm stil not sure if I should merge your changes with my rewrite,
since that code seems a lot clearer to me.
https://github.com/steffengy/zephir/commit/f50e6b75c028d6de3cf5cd6289ae298e650a3e36
What do you think?

sergeyklay:
@steffengy I think your changes again will break backwards compatibility because actually `zephir_camelise` didn't normalize string, as you do

sergeyklay:
@steffengy What happened with Appveyor?

steffengy:
@sergeyklay 
seems like a connection-drop to github.

The only thing that actually changes is that something like `__test` wouldn't get transformed to `_test` but to `Test`, which is actually what you'd expect.

Is there anything in phalcon relying on the old malicious behavior in these rare cases?

sergeyklay:
@steffengy No, actually nothing in phalcon not rely on this behavior. But we have public api here https://github.com/phalcon/cphalcon/blob/master/phalcon/text.zep#L47

sergeyklay:
@steffengy Well, I like your solution :) What solution do you think should be merged?

virgofx:
My personal opinion is that the character mask is more appropriate than the `alt` variable, provided that `strchr` is fast enough to not impact performance.

Also can we update the public API on the `Phalcon\Text` to include the extra parameters in docblock?

steffengy:
@sergeyklay I'll take a look at how much relies on that and will discuss with @andresgutierrez soon.

sergeyklay:
@virgofx Yes. We'll do it

sergeyklay:
@steffengy What do you think #1090

steffengy:
@sergeyklay 
What is with #1090, does by any chance your or my solution fix that (or is it already fixed)?
Andres seems to have tagged 0.9.2 now, so that we can decide for either solution,
so we should choose the "most-working" one.

sergeyklay:
@steffengy Yeah, creating the 0.9.2 tag a great solution that allows to return the old behavior. Indeed, we can now choose the best solution. It seems to me the #1090 can be fixed by removing https://github.com/sergeyklay/zephir/blob/3b2a283d881ca811f1717ef09ddf8a2412c3b89b/kernels/ZendEngine2/string.c#L304

steffengy:
@sergeyklay 
that line is surely elementary.
have you checked if my rewritten solution works for that?

sergeyklay:
@steffengy No, if you are talking about concrete testing, I didn't do it. But at first glance looks good to me. You can get current (un)camelize tests and run them

PR #1261, opened 2016-05-27
Fix #1260, Added stripcslashes support for PHP7

### LINKED ISSUE #1260: [php7] stripcslashes doesnt work/ causes seg
test.zep

``` zep
namespace Test;

class Test {

    public function koko(value) {

        return stripcslashes(value);
    }
}
```

``` php
<?php

echo "<pre>";

use Test\Test;

$a = new Test();

$a->koko("yep");
```

output from console: 

`
<pre>php: symbol lookup error: /opt/php-7004/lib/php/extensions/no-debug-non-zts-20151012/test.so: undefined symbol: zephir_stripcslashes
`

in browser its 500 ...

ovr:
Hey!
- [X] Fix
- [X] Tests

Thanks

### DIFF
--- a/kernels/ZendEngine3/string.c
+++ b/kernels/ZendEngine3/string.c
@@ -1326,3 +1326,23 @@ void zephir_stripslashes(zval *return_value, zval *str)
 		zval_dtor(&copy);
 	}
 }
+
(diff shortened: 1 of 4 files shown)


andresgutierrez:
Thanks

PR #1262, opened 2016-05-27
[Composer] Use latest phpcs

ovr:
Hey!
Thanks


ovr:
ping @andresgutierrez

PR #1265, opened 2016-06-11
Add support for module globals:

hylent:
When some global value in config.json has option `module` set to true, it will NOT be reinited per request.


sergeyklay:

steffengy:
+1 from me, @andresgutierrez any objection?

andresgutierrez:
Thanks

PR #1266, opened 2016-06-13
Fix risky unit tests

hervems:
- Migration to PHP Unit 5.4 and more by fixing Risky unit tests.
- Possibility to use the loader defined by Zephir without any warning.
- Activation of PHP Unit Coverage by adding filter section in phpunit.xml.dist
- Possiblity to use phpdbg to have the coverage
- Add some unit tests on ConfigException / Utils

### DIFF
--- a/unit-tests/Extension/ExitDieTest.php
+++ b/unit-tests/Extension/ExitDieTest.php
@@ -23,16 +23,12 @@ class ExitDieTest extends \PHPUnit_Framework_TestCase
 {
     public function testExitDie()
     {
-        /* Only available since PHP 5.4 */
-        if (defined('PHP_BINARY')) {
-            $phpBinary = constant('PHP_BINARY');
-            /* If we use phpdbg, you need to add options -qrr */
-            if (defined('PHP_SAPI') && constant('PHP_SAPI') == 'phpdbg') {
-                $phpBinary .= ' -qrr';
-            }
-        } else {
-            $phpBinary = PHP_BINDIR . DIRECTORY_SEPARATOR . 'php';
+        $phpBinary = constant('PHP_BINARY');
+        /* If we use phpdbg, you need to add options -qrr */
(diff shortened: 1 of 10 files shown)


sergeyklay:
Looks good to me

sergeyklay:
@hervems Could you please fix build?

sergeyklay:
@hervems  ping

unit-tests/Extension/ExitDieTest.php
```diff
@@ -23,9 +23,13 @@ class ExitDieTest extends \PHPUnit_Framework_TestCase
 {
     public function testExitDie()
     {
-        /* Only available since PHP 5.4*/
+        /* Only available since PHP 5.4 */
         if (defined('PHP_BINARY')) {
```
sergeyklay: @hervems It seems to me this may be cleaned safely due to the fact that latest Zephir does not support PHP < 5.4

sergeyklay:
@steffengy Could you please sort out with appveyor issue?

hervems:
For appveyor build, I think the problem is that PHP_EOL is not \r\n
while the var_dump uses a correct end of line ("\r\n").
I can try to do a hack, but I can't test the result because I don't have Microsoft Visual Studio Compiler (cl.exe). Do you have any idea to have a environment in local?

sergeyklay:
@hervems How about something like this https://hub.docker.com/r/ambakshi/wine-x11-vcpython27/

steffengy:
You actually do not need a full visual studio since quite a while, but only a rather small part of it available here:
http://landinghub.visualstudio.com/visual-cpp-build-tools

sergeyklay:
@steffengy Could you please take a look?

hervems:
All builds are ok (travis + appveyor).
By my side, I tested on Windows 7 with Visual Studio 11.

andresgutierrez:
Thanks

PR #1271, opened 2016-06-17
Update let.c

ghost:
can`t run test


sergeyklay:
runtime is at a very experimental stage and nowhere near usable.

@steffengy

steffengy:
runtime is not supported anymore and very outdated.

PR #1277, opened 2016-07-03
Fixed generating params prototype

sergeyklay:
**Zephir**

``` zephir
public function functionName(array! someVar)
{
    var_dump(someVar);
}
```

```
$ zephir stubs
```

**Before**

``` php
public function functionName($someVar) {}
```

**Now**

``` php
public function functionName(array $someVar) {}
```


sergeyklay:

sergeyklay:
@steffengy https://github.com/phalcon/phalcon-devtools/pull/705 👍

PR #1283, opened 2016-07-25
Fixed phpize and sudo

sergeyklay:
1. In some cases `phpize` is called `phpize5`, besides in some cases it should be used as `/usr/bin/phpize` instead of only `phpize`.
2. In some systems (for example docker containers) there is no `sudo` by default.
3. Removed not needed ocular from Zephir build


andresgutierrez:
👍

Issue #1289, opened 2016-08-08
can remove the --silent flag?


ucando:
I use zephir in FreeBSD, and its make can't support the --silent flag. I must remove this in Library/Compiler.php so that it can run correctly.

sergeyklay:
Fixed in the `master` branch.

PR #1292, opened 2016-08-10
Added Utils::isBsd, Utils::isMacOs

sergeyklay:
The part of: #1289


sergeyklay:
@andresgutierrez Can you please review?

PR #1293, opened 2016-08-11
Adapted make to use in BSD systems

sergeyklay:
Refs: #1289

**FreeBSD**
`man make`

```
-s       Do not echo any commands as they are executed.  Equivalent to
         specifying `@' before each command line in the makefile.
```

**Linux**
`man make`

```
-s, --silent, --quiet
Silent operation; do not print the commands as they are executed.
```


sergeyklay:
@andresgutierrez Could you please review?

PR #1308, opened 2016-08-30
Fixed PHP7 (ZendEngine3) zephir_array_isset_string_fetch, zephir_array_isset_long_fetch when source and destination pointers are equal

allflame:
Problem arises when you write something like fetch a, a[key] using zephir.
In`zephir_array_isset_string_fetch` and `zephir_array_isset_long_fetch` is_null flag will be set on source zval internal structure (since source and destination point to the same memory address) before evaluating `likely(Z_TYPE_P(arr) == IS_ARRAY)` which will result in failing the condition and return 0
Check ZendEngine2/array.c for comparison.

Will resolve  phalcon/cphalcon#12147


andresgutierrez:
Thanks, can you please point me to one of the lines where this bug does affect Phalcon?

allflame:
@andresgutierrez 
https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model/resultset/simple.zep#L204
https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model.zep#L3630

andresgutierrez:
Thanks

PR #1312, opened 2016-09-02
Fixed Zephir\Call::addCallStatusOrJump

sergeyklay:
`$compilationContext->currentTryCatch` instead `$compilationContext->insideTryCatch`

Refs: https://github.com/phalcon/cphalcon/issues/12154#issuecomment-244155923


sergeyklay:
@andresgutierrez Could you please take a look

Issue #1192, opened 2016-02-28
Zephir does not know about parent methods


sergeyklay:
I noticed that the Zephir does not know anything about object's parent methods.

**How to reproduce:**

```
$ pwd
/home/klay/projects/zephir/phalcon/cphalcon
```

```
$ git br
  2.0.x
* 2.1.x
  master
```

```
$ git fetch --all
$ git rebase origin/2.1.x
```

```
$ zephir build --export-classes
```

```
$ php -r 'var_dump(method_exists(new Phalcon\Logger\Adapter\File("/tmp/a.log"), "log"));'
bool(true)
```

Lets test Zephir

**config.json**

```
    ...

    "external-dependencies": {
            "phalcon": "/home/klay/projects/zephir/phalcon/cphalcon"
    }

    ...

```

``` zephir
namespace Test;

use Phalcon\Logger;
use Phalcon\Logger\Adapter\File;

class A
{
    public function __construct()
    {
        var logger;

        let logger = new File("/tmp/the-log.log");

        logger->log("This is a message");
        logger->log("This is an error", Logger::ERROR);
        logger->error("This is another error");
        logger->close();
    }
}

```

```
$ zephir build
Zephir\CompilerException: Class 'Phalcon\Logger\Adapter\File'
does not implement method: 'log' in /tmp/z/test/test/a.zep on line 14

            logger->log("This is a message");
    ----------------------------------------^
```

andresgutierrez:
This is probably only happening on external classes

sergeyklay:
Refs #304

Izopi4a:
and probably same as my issue here : https://github.com/phalcon/zephir/issues/1191

sergeyklay:
@steffengy Could you please deal with this

sergeyklay:
@mervick Thanks

Issue #304, opened 2014-03-29
Use phalcon from zephir module


DaveM2011:
As the title states how can i use phalcon from within my own zephir module

```
namespace Rapid\Models;
class Users extends \Phalcon\Mvc\Model
{
}
```

Thanks Dave

DaveM2011:
I did make a little progress making it read the model class but then it fails with 

```
Zephir\CompilerException: Method 'Rapid\Model\Tokens::belongsTo' called with a w
rong number of parameters, the method has: 0, passed: 4 in /root/rapid/rapid/mod
el/tokens.zep on line 29

this->belongsTo("user_id", "\\Rapid\\Model\\Users", "id", ["alias": "User", "foreignKey": true]);
-----------------------------------------------------------------------------------------------^

```

MethodCall.php Line: 209

I have no idea how to fix it ive tried adding headers, adjusting the bottom of ClassDefinition.php adding

```
case 'phalcon\mvc\model':
                $compilationContext->headersManager->add('ext/phalcon/phalcon/mvc/model');
                $classEntry = 'phalcon_mvc_model_ce';
                break;
```

Still no luck :(

DaveM2011:
OMG thats great i will try your fix shortly i have to take a look at other software most of today but should be able to test it later on and ill let you know :)

> For every phalcon class you want to extend you will have to find it's header file and add PHPAPI right after extern and before zend_class_entry

i wondered why PHPAPI was not in the header files i thought maybe zephir added it elsewhere :)

OpenEX:
Based on what has been already said here, i've been thinking about making my Zephir AaaaaaController class visible from PHP with Phalcon on board. All compiles great (installed/ini/extension order), but the class is not visible when i write http://domainwiththeapp/aaaaaa  . What i wouldn't like to do is to create a class in PHP extending my AaaaaaController class - this i did with success. . I guess the problem is with namespace. Is there a way to make this class visible for this address and for phalcon. Any suggestions would be appreciated.

vexor-dev:
<code>zephir build --export-classes</code>

don't work to compile phalcon anymore (testet with branches 2.0.0, 2.0.x and 2.1.x and Zephir 0.7.1b) . Compilation stops with:

<code>Zephir\CompilerException: updateArray: variable</code>

Is there a equivalent for  --export-classes  in:
<code>
cd cphalcon/build
</code>
<code>
./install
</code>

steffengy:
[...]

PR #1319, opened 2016-09-09
Fixed phalcon usage as module in zephir

mervick:
Fixed usage the external zephir extensions.
Fixed getting/checking methods/properties/constants for the parent classes of external extensions.

fix https://github.com/phalcon/zephir/issues/304
fix https://github.com/phalcon/zephir/issues/1192

### DIFF
--- a/Library/ClassDefinition.php
+++ b/Library/ClassDefinition.php
@@ -138,6 +138,11 @@ class ClassDefinition
     protected $isGeneratedConstructor = false;
 
     /**
+     * @var Compiler
+     */
+    protected $compiler;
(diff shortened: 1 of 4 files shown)


andresgutierrez:
Please don't introduce new static methods, you can use compilation context to pass global state

sergeyklay:
@mervick Could you please fix it?

mervick:
@sergeyklay I guess I can

mervick:
Removed static method, uses compilation context

sergeyklay (approved):

PR #1322, opened 2016-09-14
Use EG(fake_scope) replaced EG(scope)

dreamsxin:


andresgutierrez:
Can you please enable tests for PHP 7.1?

sergeyklay (approved):

dreamsxin:
OK, now just compiled through for PHP7.1, after add it.

andresgutierrez:
Thanks

Issue #1323, opened 2016-09-14
Better Support FreeBSD


josh4trunks:
I was working to get this ported to FreeBSD. There are some issues I hit that I can work around but would be nice to have addressed.
- [x] Use `#!/usr/bin/env bash` instead of `#!/bin/bash`
  - FreeBSD never installs packages to /bin, and bash is installed as a package since it is not the default shell.
- [x] Make 'bin/zepher' completely POSIX compliant to drop the dependency for **bash**
  - This might not be doable in a clean way so we may need to skip this.
- [x] Make 'parser/parser/build_linux.sh' completely POSIX compliant to drop the dependency for **bash**
- [x] Execute 'parser/parser/build_linux.sh' using /bin/sh instead of directly so this script doesn't necessarily need execute permissions.
- [ ] Replace instances of "gcc" with "cc" to drop the dependency for **gcc**
  - When I tested compiling an extension without **gcc** installed it failed because gcc was missing. I might be able to work around this by setting the CC variable to cc, but I did not have time to test this.
- [ ] Drop dependency for sudo
  - It would be nice to allow systems admins, that do not have **sudo** installed to still install extensions. I noticed when I built an extension as root, without sudo installed, the extension was silently not installed.
  - Maybe this could be accomplished with **su**? If not, we can just disregard this.
##

josh4trunks:
I had a possible idea to tackle the **sudo** point. Maybe we could do a check on the current UID and only  use sudo if it is not '0'?

sergeyklay:
@josh4trunks I like this approach

josh4trunks:
Is there any reason https://github.com/phalcon/zephir/blob/master/Library/Compiler.php and https://github.com/phalcon/zephir/blob/master/parser/parser/build_linux.sh have `gcc` / `CC="gcc"` instead of `cc`?

sergeyklay:
@andresgutierrez

josh4trunks:
Any word on the the need for GCC?

It would be great to get this working by the next zephir release. That way we could add a stable version to the ports tree. See the request here https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=211144

sergeyklay:
@josh4trunks Any progress?

josh4trunks:
I submitted the port for review and no FreeBSD ports reviewer has looked at it =\
I could try authoring a new bug report (since I did not start that one) maybe that would get more attention.

sergeyklay:
[...]

PR #1324, opened 2016-09-14
Better Support FreeBSD

josh4trunks:
This addresses the 1st, 2nd, 3rd, and 4th points here https://github.com/phalcon/zephir/issues/1323.

### DIFF
--- a/Library/Compiler.php
+++ b/Library/Compiler.php
@@ -263,7 +263,7 @@ class Compiler
                     $oldCwd = getcwd();
                     chdir($currentDir);
 
-                    echo shell_exec('cd parser && ./build_linux.sh');
+                    echo shell_exec('cd parser && /bin/sh ./build_linux.sh');
                     exec('make clean && phpize --clean', $output, $exit);
                     $this->logger->output('Preparing for parser compilation...');
                     exec('phpize', $output, $exit);
--- a/bin/zephir
+++ b/bin/zephir
@@ -19,7 +19,7 @@ if [ -z "$ZEPHIRDIR" ]; then
   fi
 fi
 
-if [ $1 && $2 && $3 && "$1"="-c" ]; then
+if [ $1 -a $2 -a $3 -a $1 = "-c" ]; then
     php -d safe_mode=Off -d enable_dl=On -c $2 $ZEPHIRDIR/compiler.php ${*:3}
 else
     php -d safe_mode=Off -d enable_dl=On $ZEPHIRDIR/compiler.php $*
--- a/bin/zephir
+++ b/bin/zephir
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 if [ -z "$ZEPHIRDIR" ]; then
   ZEPHIRDIR="%ZEPHIRDIR%"
   if [[ "$ZEPHIRDIR" == "%""ZEPHIRDIR""%" ]]; then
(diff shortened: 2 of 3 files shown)


josh4trunks:
To note, I have not had a chance to test this. I'm just making the changes based on documentation.

I can test this on FreeBSD soon, it would also be good if we could get this tested on Linux.

sergeyklay (approved):

josh4trunks:
Ok, tested this as working on FreeBSD. Do you mind testing on linux?

josh4trunks:
@sergeyklay this is ready for testing. do you mind reviewing and commenting?

Thanks

sergeyklay:
@josh4trunks Thanks for contributing. I'll test as soon as I can.

sergeyklay (commented):
In general looks good to me

Library/Compiler.php
```diff
@@ -263,7 +263,7 @@ public function compileParser()
                     $oldCwd = getcwd();
                     chdir($currentDir);
 
-                    echo shell_exec('cd parser && ./build_linux.sh');
+                    echo shell_exec('cd parser && /bin/sh ./build_linux.sh');
```
sergeyklay: How about 
```php
echo shell_exec('cd parser && $(which sh) ./build_linux.sh');
```

sergeyklay (approved):

josh4trunks:
I saw you previous comment about `$(which sh)` and that worked for me as well. I guess you could also just start the script with `#!/usr/bin/env sh` to use the environment default.

Either way, these changes work well in my testing.

josh4trunks:
Anything else needed for this to be merged?

sergeyklay:
Thanks

josh4trunks:
nice, thank you

bin/zephir
```diff
@@ -19,9 +19,8 @@ if [ -z "$ZEPHIRDIR" ]; then
   fi
 fi
 
-if [[ $1 && $2 && $3 && "$1"=="-c" ]]; then
+if [ $1 -a $2 -a $3 -a $1 = "-c" ]; then
```
sergeyklay, line 22: @josh4trunks 

``` sh
zephir fullclean && zephir build --backend=ZendEngine3
/usr/local/bin/zephir: 22: [: -a: unexpected operator
```

Are you sure about `#!/bin/sh` and `-a` ?

josh4trunks: I don't think /bin/sh matters here. What matters is I changed it to `[` which is the same as executing `test`. Here's the documentation for **test** on FreeBSD.
https://www.freebsd.org/cgi/man.cgi?test

I would think it work the same on any system because I see the mention of POSIX, but then I noticed this sentence...
"In particular, only expressions containing -a, -o, ( or ) can be ambiguous."

I can check this on my arch linux system at home. I wonder if this is what I should be checking as well.
https://linux.die.net/man/1/test

josh4trunks: just a guess, but maybe your version of test has some of those variables empty and is essentially running `test -a`

maybe adding a ' around each variable would work for you?
`if [ '$1' -a '$2' -a '$3' -a '$1' = "-c" ]; then`
I haven't had a chance to check this yet though

sergeyklay: https://github.com/phalcon/zephir/pull/1332

josh4trunks:
I think the issue was, if $1 was empty, test ignored it and saw -a as the first argument. Your solution, or using parenthesis would fix this.

PR #1329, opened 2016-09-25
Fix deprecated phpunit method / delete assert usage / use phpunit assertion

hervems:
- Fix deprecated phpunit method setExpectedException (deprecated since PHP Unit 5.2.0)
- Change the usage of assert functions to PHP Unit assertion methods.
- Delete the var_dump line in PregmatchTest class. This is the NULL value in the result of phpunit with PHP < 7.
- Use assertSame rather than assertEquals when it's possible.
- Use assertSame rather than assertTrue.


sergeyklay (approved):

sergeyklay:
Thanks

PR #1332, opened 2016-09-29
Avoid using "-a" or "-o" with "["

sergeyklay:
Since POSIX doesn't define the behavior of `[` with complex sets of tests, we should avoid using `-a` or `-o` with `[`.

Refs: [Conditional Blocks](http://mywiki.wooledge.org/BashGuide/TestsAndConditionals?highlight=%28conditional%29#Conditional_Blocks_.28if.2C_test_and_.5B.5B.29)

### DIFF
--- a/bin/zephir
+++ b/bin/zephir
@@ -19,7 +19,7 @@ if [ -z "$ZEPHIRDIR" ]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


josh4trunks:
this works for me on FreeBSD

PR #1340, opened 2016-10-16
Updating Compiler To Use Extension-Name For so file name.

Fenikkusu:
If you use the config option 'extension-name', the file created still uses the 'namespace' config option. This updates the compile to use the extension-name instead with a fallback to the namespace.

Note: Didn't see any tests for the compiler.


sergeyklay:
Thanks

Issue #1339, opened 2016-10-13
create_instance_params seg fault when more than 10 params


Jurigag:
create_instance_params is seg faulting when creating class with more than 10 dependencies:

```
#0  0x00007ffff393957b in _zend_is_inconsistent (ht=0xa, file=0x7ffff3f0cb58 "/root/.phpbrew/build/5.6.19-debug/Zend/zend_hash.c", line=1044)
    at /root/.phpbrew/build/5.6.19-debug/Zend/zend_hash.c:60
#1  0x00007ffff393c5e9 in zend_hash_move_forward_ex (ht=0xa, pos=0x7fffffffb568) at /root/.phpbrew/build/5.6.19-debug/Zend/zend_hash.c:1044
#2  0x00007fffdcfdfa44 in zephir_create_instance_params () from /root/.phpbrew/php/5.6.19-debug/lib/php/extensions/debug-non-zts-20131226/instancetest.so
#3  0x00007fffdcffee54 in zim_Instancetest_ExampleService_test () from /root/.phpbrew/php/5.6.19-debug/lib/php/extensions/debug-non-zts-20131226/instancetest.so
#4  0x00007ffff396e6dd in zend_do_fcall_common_helper_SPEC (execute_data=0x555555ceabf0) at /root/.phpbrew/build/5.6.19-debug/Zend/zend_vm_execute.h:558
#5  0x00007ffff396eeae in ZEND_DO_FCALL_BY_NAME_SPEC_HANDLER (execute_data=0x555555ceabf0) at /root/.phpbrew/build/5.6.19-debug/Zend/zend_vm_execute.h:693
#6  0x00007ffff396dd4c in execute_ex (execute_data=0x555555ceabf0) at /root/.phpbrew/build/5.6.19-debug/Zend/zend_vm_execute.h:363
#7  0x00007ffff396ddd5 in zend_execute (op_array=0x555555d23b60) at /root/.phpbrew/build/5.6.19-debug/Zend/zend_vm_execute.h:388
#8  0x00007ffff392914a in zend_execute_scripts (type=8, retval=0x0, file_count=3) at /root/.phpbrew/build/5.6.19-debug/Zend/zend.c:1341
#9  0x00007ffff388f1f3 in php_execute_script (primary_file=0x7fffffffdc30) at /root/.phpbrew/build/5.6.19-debug/main/main.c:2610
[...]

PR #1342, opened 2016-10-18
Fix create_instance_params for more than 10 params for php 5.6

Jurigag:
Fixes https://github.com/phalcon/zephir/issues/1339 for php 5.6, there is still issue for php 7 so don't close this issue.


steffengy:
Thanks, I don't quite get how it's already fixed in the PHP7 version, 
since the PHP5 version is older and the version the PHP7 version is based on.

PR #1344, opened 2016-10-23
Fix ability use vendor namespace the same as external extension

mervick:
Fix throws exception "Class '$className' must be located at '$filePath' but this file is missing"
when class located in other path, for example
I have installed `phalcon/incubator`,
and compiler say me "Class 'Phalcon\Mailer\Manager' must be located as '/opt/cphalcon/phalcon/mailer/manager.zep' but this file is missing",
but class located in the other path

### DIFF
--- a/Library/Compiler.php
+++ b/Library/Compiler.php
@@ -378,6 +378,7 @@ class Compiler
      *
      * @param string $className
      * @param string $location
+     * @return boolean
      * @throws CompilerException
      * @throws Exception
      * @throws ParseException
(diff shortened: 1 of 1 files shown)


sergeyklay (approved):

sergeyklay:

steffengy:
what error message does this cause when something from an external extension
cannot be found in any path?

mervick:
If class not found in any path, when throws exception:
ReflectionException: Class {classname} does not exist
 in {path}/{filename}.zep on line {line}

selection_169

PR #1351, opened 2016-11-18
Updated method Zephir\ClassDefinition::buildFromReflection

dreamsxin:
When `config.json` set `extension-name` will exec:
```shell
configure --enable-{$extension-name}
```

### DIFF
--- a/Library/Compiler.php
+++ b/Library/Compiler.php
@@ -1197,6 +1197,10 @@ class Compiler
          * Get global namespace
          */
         $namespace = str_replace('\\', '_', $this->checkDirectory());
+        $extensionName  = $this->config->get('extension-name');
+        if (empty($extensionName) || !is_string($extensionName)) {
+            $extensionName = $namespace;
+        }
         $needConfigure = $this->generate($command);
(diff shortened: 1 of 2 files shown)


Library/Compiler.php
```diff
@@ -1197,6 +1197,10 @@ public function compile(CommandInterface $command, $development = false)
          * Get global namespace
          */
         $namespace = str_replace('\\', '_', $this->checkDirectory());
+		$extensionName  = $this->config->get('extension-name');
+		if (empty($extensionName) || !is_string($extensionName)) {
+			$extensionName = $namespace;
+		}
```
sergeyklay: > Spaces must be used to indent lines; tabs are not allowed

sergeyklay (changes requested):

Library/Compiler.php
dreamsxin: Oh, I will correct it, ths.

dreamsxin:
@sergeyklay It'sfix.

sergeyklay:
Thanks

PR #1391, opened 2017-02-06
Fixes #1387

### LINKED ISSUE #1387: PHP 7.1.1 unpack function
In php 7.1.1 there is bug in `unpack` function required number of parameters in reflection api required which is wrongly reported as 3 instead of 2. There should be added additional check for 7.1.1 php version in https://github.com/phalcon/zephir/blob/master/Library/FunctionCall.php and check it against 2 parameters.

This will prevent to compile any zephir code having unpack function.

Related:
https://www.reddit.com/r/PHP/comments/5q3kx9/why_php_changes_api_stuff_in_bugfix_version/dcw6f4n/

Jurigag:
Fixes https://github.com/phalcon/zephir/issues/1387 issue

### DIFF
--- a/Library/FunctionCall.php
+++ b/Library/FunctionCall.php
@@ -116,8 +116,17 @@ class FunctionCall extends Call
                  * Check if the number of parameters
                  */
                 $numberParameters = count($expression['parameters']);
-                if ($numberParameters < $reflector->getNumberOfRequiredParameters()) {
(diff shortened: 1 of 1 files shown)


Jurigag:
What is this appveyor error? @sergeyklay

sergeyklay:
Thanks

PR #1405, opened 2017-03-18
Fix wrong pointer type

sjinks:
fcall.c:

```c
fcic.called_scope = obj_ce ? obj_ce : (EG(current_execute_data) ? Z_OBJ(EG(current_execute_data)->This) : NULL);
```

`called_scope` is `zend_class_entry*`, `Z_OBJ`'s result is `zend_object*`.


sergeyklay:
Thank you

PR #1408, opened 2017-03-19
Fix #1392

### LINKED ISSUE #1392: PHP 7.1 get_called_class
Zephir on php 7.1 returns wrong value from get_called_class. Script to reproduce:

```zep
abstract class Test
{
    public static function count()
    {
        var_dump(get_called_class());
    }
}
```

```zep
class Xyz extends Test
{
}
```

```php
Xyz::count();
```

actual "Test"
expected "Xyz"

---

Refs: #1390, #1389,  #1372

Jurigag:


sergeyklay:
@sjinks @dreamsxin @ovr Could you please take a view

kernels/ZendEngine3/extended/fcall.c
```diff
@@ -278,7 +278,11 @@ static int zephir_is_callable_check_class(const char *name, int name_len, zend_f
 		} else {
 			fcc->called_scope = called_scope;
 			if (!fcc->object) {
-				fcc->object = Z_OBJ(EG(current_execute_data)->This);
+				#if PHP_VERSION_ID >= 70100
+					fcc->object = zend_get_this_object(EG(current_execute_data));
```
ovr, line 282: Original `zend_function_call` in PHP 7.1 are fixing `EG(current_execute_data)->This` without `fcc->object` and this thing can be "null", I think `fcc->object` is needed only for callable

Jurigag: Just in php 7.1 you can't use code I just changed it to this, check difference between those functions in php 7 and 7.1

Jurigag:
Also it's not fixing everything, there are still issues in phalcon tests. I need to investigate it more, maybe somewhere else more changes

sergeyklay:
Fixed in https://github.com/sergeyklay/issue1394/pull/6

PR #1419, opened 2017-03-24
Use ZVAL_COPY_VALUE when passing parameters

sjinks:


kernels/ZendEngine3/fcall.c
```diff
@@ -283,8 +283,7 @@ int zephir_call_user_function(zval *object_pp, zend_class_entry *obj_ce, zephir_
 
 	zval p[fci.param_count];
 	for (i=0; i<fci.param_count; ++i) {
-		/** FIXME: ZVAL_COPY or ZVAL_COPY_VALUE? */
-		ZVAL_COPY(&p[i], params[i]);
+		ZVAL_COPY_VALUE(&p[i], params[i]);
```
sergeyklay, line 286: You have a good memory :smiley:

sjinks: Just experience :-) I had a similar issue in PHP-CPP

sjinks: "Parameter 3 to preg_match_all() expected to be a reference, value given" - this was the clue

sergeyklay:
Is it ready to merging?

sergeyklay:
Good catch! Thank you 👍

PR #1421, opened 2017-03-25
Added clang support

sergeyklay:


.travis.yml
```diff
@@ -15,6 +11,20 @@ php:
 git:
   depth: 1
 
+compiler:
```
sjinks, line 14: Does it work this way? Last time I checked, Travis ignored "compiler" if language: php is set

sergeyklay: Works fine. See https://github.com/phalcon/zephir/blob/master/.travis.yml#L68

PR #1428, opened 2017-03-26
Some test for require

Jurigag:
This is test for https://travis-ci.org/phalcon/cphalcon/jobs/215182360#L2132

### DIFF
--- a/test/requires.zep
+++ b/test/requires.zep
@@ -1,8 +1,11 @@
 
 namespace Test;
 
+use Test\Requires\External3;
+
 class Requires
 {
+	protected content;
 
 	public function requireExternal1(var path)
 	{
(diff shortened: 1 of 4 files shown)


Jurigag:
@sjinks well this is correctly creating same issue as in phalcon repository, though for some reason error message is different than in phalcon though i tried to make a same code.

For first i thought maybe someVariable needed to be initialized with empty array, like in phalcon repository in plain php there were some error when tried adding new array elementy to property defined as string and i thought it was related but it's not a caused, even when change someVariable to something like `protected someVariable = [];` problem still happens.

Jurigag:
@sjinks just to make sure - i checked the similar code with plain php and it works fine without issue so it's not some php 7.1 change

Issue #251, opened 2014-03-04
Private methods can't be called by owning class from parent


damienalexandre:
Hi there :neckbeard: 

I'm trying to port this PHP code to Zephir:

``` php
<?php

class Foo extends Bar {
    public function run() {
        $this->runBar();
    }
}

class Bar {
    public function runBar() {
        $this->isPrivate();
    }

    private function isPrivate() {
        echo "YEAH";
    }
}

$f = new Foo(); 
$f->run();
```

So here is my Zephir code:

```zep
namespace Jolicoucou;

class Foo extends \Jolicoucou\Bar
{
    public function run()
    {
      this->runBar();
    }
}
```

```zep
namespace Jolicoucou;

class Bar
{
    public function runBar()
    {
      this->isPrivate();
    }

    private function isPrivate()
    {
      echo "IS NEVER CALLED";
    }
}
```

And the error I get:

```
php > $f = new Jolicoucou\Foo(); $f->run();
PHP Warning:  Invalid callback Jolicoucou\Foo::isprivate, cannot access private method Jolicoucou\Foo::isPrivate() in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. Jolicoucou\Foo->run() php shell code:1
PHP   3. Jolicoucou\Bar->runBar() php shell code:1
PHP Fatal error:  Call to undefined function Jolicoucou\Foo::isprivate() in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. Jolicoucou\Foo->run() php shell code:1
PHP   3. Jolicoucou\Bar->runBar() php shell code:1
```

A class owning a private method can't call this method if we use a child class? Running `$f = new Jolicoucou\Bar(); $f->runBar();` works fine.

Thanks a lot,
:beer: Damien

sjinks:
Right now only PHP 5.4, 5.5, and 7.0 is affected

pentagonal:
# **PHP-7.0**

Test using `Pimple\Container` implements `ArrayAccess`
Extended by `Slim Container` class. (Affected By Magic Method)
![image](https://user-images.githubusercontent.com/17354203/33227027-0ab9c868-d1cd-11e7-8e64-01e81d274923.png)

![image](https://user-images.githubusercontent.com/17354203/33227033-266ba6da-d1cd-11e7-80d4-737d3348f356.png)

Looks like the the code has jumping...
The magic method called first than constructor.
[...]

PR #1429, opened 2017-03-26
Test calling protected method

Jurigag:
On phalcon repo there is some weird code that protected method of some object is called without any problems from other object, on php 7.1 it's started to failing(**IT SHOULD ON ALL VERSIONS**) so this is testing this.

### DIFF
--- a/test/mcall.zep
+++ b/test/mcall.zep
@@ -256,6 +256,14 @@ class Mcall
 		return a->setCallable([this, "bb"]);
 	}
 
+	public function callProtectedFromOther()
+	{
+		var a;
+		let a = new Oo\OoOther();
+
+		return a->callProtected(this);
+	}
+
 	public function aa()
 	{
 		var a;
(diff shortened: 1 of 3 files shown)


sjinks:
Related: #251 (the reason for the bug is the same)

sjinks:
Related: #1507 (the fix should be ported to ZE2 kernel)

sergeyklay:
Calling parent's private methods from the child's public ones doesn't provided for PHP < 5.6. This feature works well now for PHP 5.6, 7.0, 7.1, 7.2 and even for 7.3 (nightly build)  For older PHP versions like PHP 5.5 this will not be implemented. I want to close this issue because I don't want a bunch of these lying around. I'd like to add support for anything/everything at some point, but keeping the issue open doesn't help that. :) If anyone wants to get started, I'd love that.

Refs:
- https://github.com/phalcon/zephir/pull/1467
- https://github.com/phalcon/zephir/pull/1429
- https://github.com/phalcon/zephir/issues/251

Issue #1336, opened 2016-10-06
Multicatch block is causing seg fault/exception in multicatch is caught two times


Jurigag:
```
0x00007fffe950b4f4 in zval_addref_p (pz=0x7fffffff9dd0) at /root/.phpbrew/php/7.0.5-debug/include/php/Zend/zend_types.h:822
822             return ++GC_REFCOUNT(Z_COUNTED_P(pz));
(gdb) bt
#0  0x00007fffe950b4f4 in zval_addref_p (pz=0x7fffffff9dd0) at /root/.phpbrew/php/7.0.5-debug/include/php/Zend/zend_types.h:822
#1  0x00007fffe951402f in zim_Phalcon_Dispatcher__dispatch (execute_data=0x7ffff1614db0, return_value=0x7fffffffa5d0) at /root/3.0.x-my/ext/phalcon/dispatcher.zep.c:1122
#2  0x00007fffe94821f7 in zephir_call_function_opt (fci=0x7fffffffa3f0, fci_cache=0x7fffffffa3c0, info=0x7fffffffa4d0, params=0x7fffffffa560) at /root/3.0.x-my/ext/kernel/extended/fcall.c:1223
#3  0x00007fffe947ca0d in zephir_call_user_function (object_pp=0x7fffffffa620, obj_ce=0x161a3a0, type=zephir_fcall_method, function_name=0x0, retval_ptr=0x7fffffffa5d0, cache_entry=0x0, 
    cache_slot=0, param_count=0, params=0x7fffffffa560, info=0x7fffffffa4d0) at /root/3.0.x-my/ext/kernel/fcall.c:494
#4  0x00007fffe947d1e0 in zephir_call_class_method_aparams (return_value_ptr=0x7fffffffa5d0, ce=0x161a3a0, type=zephir_fcall_method, object=0x7fffffffa620, 
    method_name=0x7fffe9a8308d "_dispatch", method_len=9, cache_entry=0x0, cache_slot=0, param_count=0, params=0x7fffffffa560) at /root/3.0.x-my/ext/kernel/fcall.c:699
#5  0x00007fffe950e83e in zim_Phalcon_Dispatcher_dispatch (execute_data=0x7ffff1614d50, return_value=0x7ffff1614d30) at /root/3.0.x-my/ext/phalcon/dispatcher.zep.c:623
[...]

PR #1447, opened 2017-03-31
Test for #1336

Jurigag:
This is test for https://github.com/phalcon/zephir/issues/1336

### DIFF
--- a/test/exceptions.zep
+++ b/test/exceptions.zep
@@ -3,6 +3,9 @@ namespace Test;
 
 class Exceptions
 {
+	public internalExceptionCallable;
+    public exceptionCallable;
+
     public function testException1()
     {
         throw new Exception("hello1");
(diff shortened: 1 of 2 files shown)


Jurigag:
This issue as i remember mostly happens because of code generation, each catch block is just generated as if when they should be imho `if elseif` @sjinks

test/exceptions.zep
```diff
@@ -80,4 +83,27 @@ class Exceptions
             throw e;
         }
     }
+
+    public function testMultiException(var returnValue, var exception) {
```
sergeyklay, line 87: Could you please follow common code style

Jurigag: Yea sure, will fix it. You mean make it `string` and `<\Exception>` type and intention above?

sergeyklay: Yes, and fix braces

sjinks:
Let's move to #1450

PR #1448, opened 2017-03-31
Do not use indirection in `zephir_read_static_property_ce` for zval*

sjinks:


unit-tests/ci/after_failure.sh
```diff
@@ -4,7 +4,7 @@ shopt -s nullglob
 export LC_ALL=C
 for i in core core.*; do
 	if [ -f "$i" -a "$(file "$i" | grep -o 'core file')" ]; then
-		gdb -q $(file "$i" | grep -oE "'[^ ']+" | sed "s/^'//g") "$i" <<EOF
+		gdb -q $(phpenv which php) "$i" <<EOF
```
sergeyklay, line 7: 👍

sergeyklay:
Good catch!

unit-tests/ci/after_failure.sh
Jurigag: Well it's actually from my PR, you posted me this on slack so i changed :D

PR #1450, opened 2017-03-31
Multiple `catch`'es

sjinks:
See #1336

### DIFF
--- a/Library/Statements/TryCatchStatement.php
+++ b/Library/Statements/TryCatchStatement.php
@@ -73,7 +73,11 @@ class TryCatchStatement extends StatementAbstract
             $codePrinter->output('if (EG(exception)) {');
             $codePrinter->increaseLevel();
 
+            $exc_var = $compilationContext->symbolTable->getTempVariableForWrite('variable', $compilationContext, $compilationContext);
+            $compilationContext->backend->copyOnWrite($exc_var, 'EG(exception)', $compilationContext);
+
             $exprBuilder = BuilderFactory::getInstance();
+            $ifs         = array();
 
             foreach ($this->_statement['catches'] as $catch) {
                 if (isset($catch['variable'])) {
@@ -85,7 +89,7 @@ class TryCatchStatement extends StatementAbstract
                     $variable = $compilationContext->symbolTable->getTempVariableForWrite('variable', $compilationContext, $compilationContext);
                 }
 
-                $compilationContext->backend->copyOnWrite($variable, 'EG(exception)', $compilationContext);
+//                 $compilationContext->backend->copyOnWrite($variable, 'EG(exception)', $compilationContext);
 
                 /**
                  * @TODO, use a builder here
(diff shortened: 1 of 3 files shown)


Library/Statements/TryCatchStatement.php
```diff
@@ -97,28 +101,34 @@ public function compile(CompilationContext $compilationContext)
...
-
-                    $ifStatement = new IfStatement($ifCheck->build());
-                    $ifStatement->compile($compilationContext);
                 }
+            }
 
-                if ($variable->isTemporal()) {
-                    $variable->setIdle(true);
-                }
+            $primaryIf = $ifs[0];
```
sjinks, line 132: Is there a cleaner way to do this?

Right now this generates something like:

```с
if (...) {
} else {
    if (...) {
    } else {
        if (...) {
            ...
        }
    }
}

File: Library/Statements/TryCatchStatement.php [hunk 2]
```diff
@@ -97,28 +101,34 @@ public function compile(CompilationContext $compilationContext)
...
-                                $exprBuilder->variable($variable->getName()),
+                                $exprBuilder->variable($exc_var->getName()),
                                 $exprBuilder->variable($class['value'])
                             )
                         )
                         ->setStatements($exprBuilder->statements()->block(array_merge(
-                            array($exprBuilder->statements()->rawC('zend_clear_exception(TSRMLS_C);')),
+                            array(
+                                $exprBuilder->statements()->rawC('zend_clear_exception(TSRMLS_C);'),
+                                $exprBuilder->statements()->rawC('ZEPHIR_CPY_WRT(&' . $variable->getName() . ', &' . $exc_var->getName() . ');'),
```
sjinks: Is there a cleaner way?

`$compilationContext->backend->copyOnWrite()` writes the code, what I need is the assignment operator as this probably won't compile on ZE2

unit-tests/Extension/ExceptionsTest.php
```diff
@@ -143,4 +144,37 @@ public function testExceptionRethrow()
             $this->assertSame(8, $e->getLine());
```
sjinks: Just wonder why line number is 8.

The test says it is 11, I thought it should be 83.

@Jurigag any idea?

Library/Statements/TryCatchStatement.php
Jurigag: Well, the best would be somehow to introduce elseif somehow obviously.

unit-tests/Extension/ExceptionsTest.php
Jurigag: https://github.com/phalcon/zephir/blob/master/test/exceptions.zep#L8 this is this eight line though as i see there in code maybe you are correct, it should be most likely https://github.com/phalcon/zephir/blob/master/test/exceptions.zep#L80 ?

sergeyklay:
Looks good to me

unit-tests/Extension/ExceptionsTest.php
sjinks: Make sense… Your test patch added 3 lines to the top of test/exceptions.zep this is why the number is 11 now

sergeyklay:
As I can see there is code generation issue

Library/Statements/TryCatchStatement.php
```diff
@@ -85,7 +89,11 @@ public function compile(CompilationContext $compilationContext)
                     $variable = $compilationContext->symbolTable->getTempVariableForWrite('variable', $compilationContext, $compilationContext);
                 }
 
-                $compilationContext->backend->copyOnWrite($variable, 'EG(exception)', $compilationContext);
+                if ($compilationContext->backend->isZE3()) {
+                    $assignExceptionVarStmt = $exprBuilder->statements()->rawC('ZEPHIR_CPY_WRT(&' . $variable->getName() . ', &' . $exc_var->getName() . ');');
+                } else {
+                    $assignExceptionVarStmt = $exprBuilder->statements()->rawC('ZEPHIR_CPY_WRT(' . $variable->getName() . ', ' . $exc_var->getName() . ');');
+                }
```
sjinks, line 97: The above is a dirty hack, but I failed to make this:

```php
$exprBuilder->statements()->let(array(
     $exprBuilder->operators()->assignVariable($variable->getName(), $exprBuilder->variable($exc_var->getName()))
))->build(),
```

or similar constructs work :-(

If someone knows how to make them work, I would prefer it to this code.

sjinks:
The code works, I am therefore merging it into master, but if someone has a better idea how to handle [this](https://github.com/phalcon/zephir/pull/1450#pullrequestreview-30360486), that would be great.

Library/Statements/TryCatchStatement.php
sergeyklay: Cc: @ovr

Issue #1449, opened 2017-03-31
Missing optimizers


Jurigag:
- [ ] extract
- [ ] compact
- [x] get_defined_vars
- [ ] parse_str
- [ ] assert
- [x] func_get_args
- [x] func_get_arg
- [x] func_num_args

Source - https://wiki.php.net/rfc/forbid_dynamic_scope_introspection

sjinks:
`get_defined_vars()` is useless in Zephir: since all methods will be compiled to C functions, there is no way for the ZE to get information about variables.

`get_defined_vars()` will return variables defined in the most recent userland function down the call stack.

sjinks:
See 4d5b616c143e45dc7a3b73d952116243b5ed2cdd

sjinks:
For the same reason `compact()` is also useless: there is no local symbol table in Zephir/C functions.

In theory, `compact()` can be implemented in the parser (because only the parser knows the current symbol table) but I am not familiar with the parser good enough to implement this.

`extract()` won't work in Zephir for the very same reason: there is no symbol table, there is nowhere to extract the variables. Moreover, when used on untrusted input, `extract` poses a security risk.

sjinks:
In ZE3 only `assert("string")` is forbidden, but it can be easily replaced with smth like `assert(eval("string"))`. But again, this is not useful in Zephir, because `eval()` won't have access to local variables as there is no symbol table.

sjinks:
`(mb_)parse_str()` with one parameter also does not make sense in Zephir, and poses a security risk.

Jurigag:
Hmmm then i think maybe we could do some optimizers anyway telling that it's not supported in zephir with reason why, cuz maybe it's not so obvious really(like it doesn't have to be dynamic call but still not working).

sjinks:
Throw an exception during compilation?

Jurigag:
Maybe? @sergeyklay what you think?

sjinks:
`get_defined_vars()` implemented in #1452

Jurigag:
Also im not sure, but the same problem exists on ZE2 too, right?

sergeyklay:
In my opinion we can just throw compilation exception during compilation

Jurigag:
Okay then, i agree.

PR #1452, opened 2017-04-01
get_defined_vars() optimizer

sjinks:
@sergeyklay please see the discussion in #1449; if you think this should be merged, please do; honestly I don't see much sense in get_defined_vars() as its behaviour is counter-intuitive in Zephir.


sjinks:
I would cherry-pick 386c85c and 3670cdf though anyway.

ovr:
LGTM, but I cannot explain who will use it in Zephir :)

sergeyklay:
Thank you

Issue #1369, opened 2016-12-14
returns value incorrect


zhiephie:
I have a code use function ``dexhex`` and ``ord``
```zep
public function strToHex(string value) -> string
{
     int i = 0;
     string ret = "";
     while (i < value->length()) {
	   let ret .= dechex(ord(value[i]));
	   let i++;
    }
   return ret;
}
```
Output:
`31313131333131313131313131343434`

In this PHP
```zep
function strToHex($string){
    $hex='';
    for ($i=0; $i < strlen($string); $i++) {
         $hex .= dechex(ord($string[$i]));
    }
    return $hex;
}
```
Output:
```
746573742073656e74656e73652e2e2e
```

dreamsxin:
@sergeyklay Error operation `ZVAL_LONG `, 

```c
        i = 0;
        ZEPHIR_INIT_VAR(ret);
        ZVAL_EMPTY_STRING(ret);
        while (1) {
                if (!(i < zephir_fast_strlen_ev(value))) {
                        break;
                }
                _1$$3 = ZEPHIR_STRING_OFFSET(value, i);
                ZEPHIR_SINIT_NVAR(_2$$3);
                ZVAL_LONG(&_2$$3, _1$$3);
                ZEPHIR_CALL_FUNCTION(&_3$$3, "ord", &_4, 61, &_2$$3);
                zephir_check_call_status();
                ZEPHIR_CALL_FUNCTION(&_5$$3, "dechex", &_6, 7, _3$$3);
                zephir_check_call_status();
                zephir_concat_self(&ret, _5$$3 TSRMLS_CC);
                i++;
        }
        RETURN_CTOR(ret);
```

The lemon rule:
```lemon
xx_array_offset(R) ::= SBRACKET_OPEN xx_index_expr(I) SBRACKET_CLOSE . {
	R = I;
}
```

@zhiephie You can test this:

```zep
public function strToHex(string value) -> string
{
     int i = 0;
     string ret = "";
     string c = "";
     while (i < value->length()) {
           c = value[i];
	   let ret .= dechex(ord(c));
	   let i++;
    }
   return ret;
}
```

zhiephie:
Work it, thanks @dreamsxin, can i close this issue?

dreamsxin:
@zhiephie My pleasure, Keep open, that's bug.

zhiephie:
Closed https://github.com/phalcon/zephir/pull/1454

PR #1454, opened 2017-04-01
Fix promotion of chars

sjinks:
See #1369 

Zend Engine does not have char type, and string characters accessed in a string by offset are also strings.

Zephir decided to introduce char type but promotes it to integer which is not compatible with PHP.

In this PR:
* chars are promoted to strings in `getReadOnlyResolvedParams()`;
* char->hex() is implemented via `zephir_string_to_hex()` builtin, as `sprintf("%X", "c")` returns 0 (`sprintf` relies on argument types).

### DIFF
--- a/Library/Types/CharType.php
+++ b/Library/Types/CharType.php
@@ -50,7 +50,7 @@ class CharType extends AbstractType
     {
         $exprBuilder = BuilderFactory::getInstance();
         $functionCall = $exprBuilder->statements()
-            ->functionCall('sprintf', array($exprBuilder->literal(Types::STRING, '%X'), $caller))
+            ->functionCall('zephir_string_to_hex', array($caller))
             ->setFile($expression['file'])
             ->setLine($expression['line'])
             ->setChar($expression['char']);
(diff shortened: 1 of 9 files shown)


sjinks:
@Jurigag if you happen to have more private tests, could you please check that this change does not break them?

Library/Types/CharType.php
```diff
@@ -50,7 +50,7 @@ public function toHex($caller, CompilationContext $compilationContext, Call $cal
     {
         $exprBuilder = BuilderFactory::getInstance();
         $functionCall = $exprBuilder->statements()
-            ->functionCall('sprintf', array($exprBuilder->literal(Types::STRING, '%X'), $caller))
+            ->functionCall('zephir_string_to_hex', array($caller))
```
ovr, line 53: Comment about for this optimizer about method

> Transforms calls to method "toHex" to sprintf('%X') call

Looks like should be changed, or removed because it's pretty simple

sergeyklay (approved):

Jurigag (approved):

PR #1455, opened 2017-04-01
Suppress unnecessary output in RequiresTest

sjinks:


test/requires/External3.zep
```diff
@@ -14,5 +14,6 @@ class External3
 		ob_clean();
 		require path;
 		requires->setContent(ob_get_contents());
+		ob_clean();
```
sergeyklay, line 17: 👍

Jurigag: Oh well my mistake, just replace `ob_get_contents()` with `ob_get_clean()`

sjinks: @Jurigag nah, the test will fail :-) I think PHPUnit is calling ob_clean() after the test and since there is no output buffer any longer, it throws an exception.

sjinks: ob_get_clean() is ob_get_contents() + ob_end_clean(), not ob_clean()

Jurigag: Oh okay. If there is difference then okay.

PR #1458, opened 2017-04-01
[Runtime] Start working on PHP 7+ support - WIP!!

ovr:


runtime/zephir.c
```diff
@@ -108,7 +108,7 @@ static void zephir_initialize_memory(zend_zephir_globals *zephir_globals_ptr TSR
 static void zephir_compile_program(zval *program TSRMLS_DC)
 {
 	HashTable           *ht = Z_ARRVAL_P(program);
-	HashPosition        pos = {0};
+	HashPosition        pos;
```
ovr, line 111: @sjinks Why it was = {0} by default? Maybe you will give a cool explanation, because I am rly interested in this

runtime/zephir.c
```diff
@@ -139,7 +139,7 @@ static void zephir_compile_program(zval *program TSRMLS_DC)
 
 	zend_hash_internal_pointer_reset_ex(ht, &pos);
 	for (
-	 ; zend_hash_get_current_data_ex(ht, (void**) &z, &pos) == SUCCESS
+	 ; z = zend_hash_get_current_data_ex(ht, &pos) != NULL
```
ovr, line 142: [Why I change]
 
In PHP 7+, `zend_hash_get_current_data_ex` accept only 2 params and return value by `position`, check should != Null (null because C lang doesnot support nullptr and NULL is (void) *0)

runtime/zephir.c
```diff
@@ -161,10 +161,9 @@ static void zephir_compile_program(zval *program TSRMLS_DC)
  */
 static void zephir_parse_file(const char *file_name TSRMLS_DC)
 {
-    char *file_name_pass = (char*) file_name;
-	char *contents;
+	char *file_name_pass = (char*) file_name;
+	zend_string *contents;
```
ovr, line 165: [Why I change]

Because PHP7 introduce `zend_string` and `php_stream_copy_to_mem` return `zend_string`

runtime/zephir.c
```diff
@@ -176,9 +175,9 @@ static void zephir_parse_file(const char *file_name TSRMLS_DC)
 		return;
 	}
 
-	if ((len = php_stream_copy_to_mem(stream, &contents, maxlen, 0)) > 0) {
+	if (contents = php_stream_copy_to_mem(stream, PHP_STREAM_COPY_ALL, 0)) {
 
-		zephir_parse_program(&return_value, contents, len, file_name, NULL TSRMLS_CC);
+		zephir_parse_program(&return_value, contents, ZSTR_LEN(contents), file_name, NULL TSRMLS_CC);
```
ovr, line 180: [Why I change]

`ZSTR_LEN` is similar to `contents->len`, (len is a field of `zend_string` structure)

runtime/zephir.c
```diff
@@ -176,9 +175,9 @@ static void zephir_parse_file(const char *file_name TSRMLS_DC)
 		return;
 	}
 
-	if ((len = php_stream_copy_to_mem(stream, &contents, maxlen, 0)) > 0) {
+	if (contents = php_stream_copy_to_mem(stream, PHP_STREAM_COPY_ALL, 0)) {
```
ovr, line 178: @sjinks What flag should I use? `PHP_STREAM_COPY_ALL` is it okey?
Thanks

runtime/kernel/exceptions.c
```diff
@@ -42,7 +42,7 @@ void zephirt_throw_exception(zval *object TSRMLS_DC){
 /**
  * Throws a zval object as exception
  */
-void zephirt_throw_exception_debug(zval *object, const char *file, zend_uint line TSRMLS_DC){
+void zephirt_throw_exception_debug(zval *object, const char *file, uint line TSRMLS_DC){
```
ovr, line 45: ping @sjinks, can you help me
Should I change `zend_uint` to `uint`, or it's wrong?
Thanks

Jurigag:
Im just wondering - could we possibly somehow add tests for runtime when and if this will be finished?

ovr:
@Jurigag Will be cool to move this project-subtree to another project, and add tests
But in near feature, after porting it to PHP7

runtime/zephir.c
sjinks: `typedef uint32_t HashPosition` in PHP 7
`typedef Bucket* HashPosition` in PHP 5

I would assign it to plain 0, not sure why `{ 0 }` was used.

Bracketed initialization also works though: http://rextester.com/GXS88393

runtime/zephir.c
sjinks: Why do we cast to `(char*)`? Is it `const char*`?

runtime/kernel/exceptions.c
sjinks: Line numbers in `zend_exception.c` are `zend_long`

Jurigag:
Also is it even worth to trying to do this runtime thing with JIT coming to php? I mean will there be any performance reason to use it, or just for fun?

To be honest i think better would be for zephir/phalcon right now to focus on new features/performance/bug fixes in zephir/phalcon itself.

ovr:
> Also is it even worth to trying to do this runtime thing with JIT coming to php? I mean will there be any performance reason to use it, or just for fun?

For Fun, it's an idea

PR #1464, opened 2017-04-02
ZEPHIR_MM_RESTORE() should be called right before zephir_exit()

### LINKED ISSUE #1381: die will cause zephir cordump under php7.0
Calling die("hello"); anywhere in zep code will cause coredump of PHP

<bountysource-plugin>
---

sjinks:
Fix #1381


sjinks:
I don't see a way how to write a test with PHPUnit for this…

sergeyklay:
Good work

PR #1466, opened 2017-04-02
Code cleanup

sjinks:
Removed unused macros, unused files, etc


.travis.yml
```diff
@@ -21,19 +21,20 @@ env:
...
 matrix:
   fast_finish: true
   allow_failures:
     - php: nightly
   include:
-    - env: CC="ccache gcc" CFLAGS="-g3 -O0 -fno-delete-null-pointer-checks -Wall -fvisibility=hidden -DZEPHIR_RELEASE=1"
+    - env: CC="ccache gcc"
       php: nightly
       compiler: gcc
-    - env: CC="clang" CFLAGS="-g3 -O0 -Wall -DZEPHIR_RELEASE=1"
```
sergeyklay, line 36: @sjinks In fact the presence of `-DZEPHIR_RELEASE` flag could help to identify potential problems at development stage. isn't it?

sergeyklay:
Fine!

.travis.yml
sjinks: No, just the opposite — `ZEPHIR_RELEASE` disables the debug code. `PHALCON_RELEASE` / `ZEPHIR_RELEASE` were used for production builds.

sergeyklay: Sill can't understand..

We have a lot of such logic:
```c
#ifndef ZEPHIR_RELEASE
zend_error(E_WARNING, "Invalid arguments supplied for memnstr in %s on line %d", file, line);
#else
zend_error(E_WARNING, "Invalid arguments supplied for memnstr()");
#endif
```

Here `ZEPHIR_RELEASE` can help at development stage. Right? Probably we have to refactor this or use another identifier..

sjinks: #if**n**def

That is, if `ZEPHIR_RELEASE` is NOT defined…

sergeyklay: Oh! I'm sorry 😐

sergeyklay: What about `-O0` ?

PR #1467, opened 2017-04-02
Fix scope issues in calls (PHP 7.0)

sjinks:
See #251 

PHP 5.4 and 5.5 still need fixing.

### DIFF
--- a/kernels/ZendEngine2/fcall.c
+++ b/kernels/ZendEngine2/fcall.c
@@ -564,12 +564,13 @@ int zephir_call_user_function(zval **object_pp, zend_class_entry *obj_ce, zephir
 	//	fcic.function_handler->op_array.line_end = 0;
 	//}
 
-	/* fcic.initialized = 0; */
-#if PHP_VERSION_ID >= 50600
-	status = ZEPHIR_ZEND_CALL_FUNCTION_WRAPPER(&fci, &fcic, info TSRMLS_CC);
-#else
-	status = ZEPHIR_ZEND_CALL_FUNCTION_WRAPPER(&fci, &fcic TSRMLS_CC);
-#endif
+	fcic.initialized = 0;
+	status = zend_call_function(&fci, &fcic TSRMLS_CC);
+//#if PHP_VERSION_ID >= 50600
+//	status = ZEPHIR_ZEND_CALL_FUNCTION_WRAPPER(&fci, &fcic, info TSRMLS_CC);
+//#else
+//	status = ZEPHIR_ZEND_CALL_FUNCTION_WRAPPER(&fci, &fcic TSRMLS_CC);
+//#endif
 
 /*
 	if (fcic.initialized && cache_entry) {
--- a/test/bar.zep
+++ b/test/bar.zep
@@ -1,0 +1,14 @@
+namespace Test;
+
+class Bar
+{
+	public function runBar()
+	{
+		return this->isPrivate();
+	}
+
+	private function isPrivate()
+	{
+		return 20170402;
+	}
+}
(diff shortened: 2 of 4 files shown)


Izopi4a:
i suggest dropping  5.4 and 5.5 support 

http://php.net/supported-versions.php

Jurigag:
Well as i remember klay said we will drop php 5 fully when phalcon 4 is released.

sergeyklay:
@sjinks @Izopi4a We don't support 5.4. The 5.5 support needed for Phalcon 3.x (LTS)

kernels/ZendEngine2/fcall.c
```diff
@@ -564,12 +564,13 @@ int zephir_call_user_function(zval **object_pp, zend_class_entry *obj_ce, zephir
 	//	fcic.function_handler->op_array.line_end = 0;
 	//}
 
-	/* fcic.initialized = 0; */
-#if PHP_VERSION_ID >= 50600
-	status = ZEPHIR_ZEND_CALL_FUNCTION_WRAPPER(&fci, &fcic, info TSRMLS_CC);
```
Jurigag, line 569: So shouldn't this `ZEPHIR_ZEND_CALL_FUNCTION_WRAPPER` be removed or it's used somewhere else?

sergeyklay: @sjinks I'm pretty sure that it doesn't resolve the issue for the PHP 7.0 (ZE3) because it is ZE2. Also it seems this may cause performance regression for PHP 5.6.

kernels/ZendEngine2/fcall.c
```diff
@@ -564,12 +564,13 @@ int zephir_call_user_function(zval **object_pp, zend_class_entry *obj_ce, zephir
...
-#else
-	status = ZEPHIR_ZEND_CALL_FUNCTION_WRAPPER(&fci, &fcic TSRMLS_CC);
-#endif
+	fcic.initialized = 0;
+	status = zend_call_function(&fci, &fcic TSRMLS_CC);
```
sergeyklay, line 568: @sjinks Just inspected this in ZE2 and ZE3.

**ZE2** (before `zend_call_function`):
```
print fcic.called_scope.name.val   # (zend_class_entry *) 0x0
print fcic.calling_scope.name.val  # (zend_class_entry *) 0x0
print fcic.object_ptr              # (zval *) 0x0
print print fcic.function_handler  # (zend_function *) 0x0
```

**ZE3** (before `zend_call_function`):
```
printf "%s\n", fcic.called_scope.name.val                     # Group\PrivateCaller
printf "%s\n", fcic.calling_scope.name.val                    # Group\PrivateCaller
printf "%s\n", fcic.object.ce.name.val                        # Group\PrivateCaller
printf "%s\n", fcic.function_handler.common.function_name.val # runBar
```

Actually I like your solution with `populate_fcic` and `resolve_callable` for ZE3. Could you please try to backport they to the ZE2?

sergeyklay:
Calling parent's private methods from the child's public ones doesn't provided for PHP < 5.6. This feature works well now for PHP 5.6, 7.0, 7.1, 7.2 and even for 7.3 (nightly build)  For older PHP versions like PHP 5.5 this will not be implemented. I want to close this issue because I don't want a bunch of these lying around. I'd like to add support for anything/everything at some point, but keeping the issue open doesn't help that. :) If anyone wants to get started, I'd love that.

Refs:
- https://github.com/phalcon/zephir/pull/1467
- https://github.com/phalcon/zephir/pull/1429
- https://github.com/phalcon/zephir/issues/251

Issue #1475, opened 2017-04-07
Memory leaks in Zephir


sjinks:
```
==32320== 
==32320== HEAP SUMMARY:
==32320==     in use at exit: 22,383 bytes in 259 blocks
==32320==   total heap usage: 313,020 allocs, 312,761 frees, 66,395,022 bytes allocated
==32320== 
==32320== 16 bytes in 1 blocks are definitely lost in loss record 13 of 70
==32320==    at 0x4C2CB3F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==32320==    by 0x105F7BA3: zephir_call_user_function (fcall.c:323)
==32320==    by 0x105F8330: zephir_call_class_method_aparams (fcall.c:509)
==32320==    by 0x106CDEA0: zim_Test_Sort_quick (sort.zep.c:95)
==32320==    by 0x848517: zend_call_function (zend_execute_API.c:885)
==32320==    by 0x105F7B1A: zephir_call_user_function (fcall.c:303)
==32320==    by 0x105F8330: zephir_call_class_method_aparams (fcall.c:509)
==32320==    by 0x106CDEA0: zim_Test_Sort_quick (sort.zep.c:95)
==32320==    by 0x8C052B: ZEND_DO_FCALL_SPEC_HANDLER (zend_vm_execute.h:842)
==32320==    by 0x8BF3AB: execute_ex (zend_vm_execute.h:414)
==32320==    by 0x8483F1: zend_call_function (zend_execute_API.c:866)
==32320==    by 0x5AF943: zim_reflection_method_invokeArgs (php_reflection.c:3348)
==32320== 
==32320== 16 bytes in 1 blocks are definitely lost in loss record 14 of 70
==32320==    at 0x4C2CB3F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==32320==    by 0x105F7BA3: zephir_call_user_function (fcall.c:323)
==32320==    by 0x105F8330: zephir_call_class_method_aparams (fcall.c:509)
==32320==    by 0x10670D83: zim_Test_Mcall_testCall7 (mcall.zep.c:242)
==32320==    by 0x8C052B: ZEND_DO_FCALL_SPEC_HANDLER (zend_vm_execute.h:842)
==32320==    by 0x8BF3AB: execute_ex (zend_vm_execute.h:414)
==32320==    by 0x8483F1: zend_call_function (zend_execute_API.c:866)
[...]

PR #1484, opened 2017-04-07
Added ability to not exit from PHPUnit test runner

sergeyklay:
Refs: #1475


sjinks:
Actually we have another issue:

https://travis-ci.org/sjinks/zephir/jobs/219949579

That is, we still need to use exit() but only if there are failures :-)

PR #701, opened 2014-12-28
Make zephir clean and fullclean works in windows.

evandropoa:

Issue #1462, opened 2017-04-01
[NFR] zephir fullclean


sergeyklay:
`zephir fullclean` should remove local cache

<img width="297" alt="screen shot 2017-04-01 at 7 36 56 pm" src="https://cloud.githubusercontent.com/assets/1256298/24580583/f77a4d74-1712-11e7-86c0-552eac13c17b.png">

sergeyklay:
Implemented in the `master` branch

PR #1503, opened 2017-04-12
Make `zephir clean` and `fullclean` work in Windows

sjinks:
See #701

### DIFF
--- a/Library/Compiler.php
+++ b/Library/Compiler.php
@@ -1040,7 +1040,13 @@ class Compiler
      */
     public function clean(CommandInterface $command)
     {
-        system('cd ext && make clean 1> /dev/null');
+        $this->fileSystem->clean();
+        if (Utils::isWindows()){
+        	system('cd ext && nmake clean-all');
+        }
+        else{
+        	system('cd ext && sudo make clean 1> /dev/null');
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Library/Compiler.php
```diff
@@ -1260,7 +1260,12 @@ public function test(CommandInterface $command)
      */
     public function clean(CommandInterface $command)
     {
-        system('cd ext && make clean 1> /dev/null');
+        $this->fileSystem->clean();
+        if (Utils::isWindows()) {
+            system('cd ext && nmake clean-all');
+        } else {
+            system('cd ext && make clean > /dev/null');
+        }
```
sergeyklay, line 1268: @sjinks Could we add something like this: https://github.com/phalcon/php-zephir-parser/blob/development/unit-tests/Extension/TestCase.php#L54-L77

related to the https://github.com/phalcon/zephir/issues/1462

sergeyklay: https://github.com/phalcon/zephir/blob/master/Library/FileSystem/HardDisk.php#L166

sjinks: Please take a look

.gitignore
```diff
@@ -94,3 +94,10 @@ parser/parser.out
 parser/Release
 parser/lemon
 parser/lemon.exe
+
+# Eclipse
+.cproject
+.project
+.settings/
+.buildpath
+.tracefile
```
sergeyklay: What about https://help.github.com/articles/ignoring-files/#create-a-global-gitignore 

See top of this file :)

sergeyklay:
Thank you!

PR #1527, opened 2017-04-27
Fix segfaults in Phalcon (ZE2)

sjinks:
Revert 61c06a1fad2a93c71d5be3367297c813689ef523
Restore the definition of `ZEPHIR_SINIT_NVAR`

### DIFF
--- a/Library/Backends/ZendEngine2/Backend.php
+++ b/Library/Backends/ZendEngine2/Backend.php
@@ -348,11 +348,6 @@ class Backend extends BaseBackend
 
         /* Initialize default values in dynamic variables */
         foreach ($variables as $variable) {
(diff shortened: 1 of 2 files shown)


Library/Backends/ZendEngine2/Backend.php
```diff
@@ -348,11 +348,6 @@ public function initializeVariableDefaults($variables, CompilationContext $compi
 
         /* Initialize default values in dynamic variables */
         foreach ($variables as $variable) {
-            if (!$this->isZE3() && $variable->isComplex() && $variable->isLocalOnly() && $variable->mustInitNull()) {
-                $codePrinter->output('zephir_memory_observe_alt(&' . $variable->getName() . ' TSRMLS_CC);');
-                $compilationContext->symbolTable->mustGrownStack(true);
-            }
-
```
sergeyklay, line 355: 👍

PR #1531, opened 2017-05-02
Added test for $_SESSION

sergeyklay:

PR #1532, opened 2017-05-07
Fix superglobals

sjinks:
See #1531

### DIFF
--- a/Library/Variable.php
+++ b/Library/Variable.php
@@ -152,13 +152,28 @@ class Variable
      *
      * @var array
      */
-    protected $complexTypes = array(
+    protected static $complexTypes = array(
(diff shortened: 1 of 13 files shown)


Library/Variable.php
```diff
@@ -1000,7 +1015,22 @@ public function observeOrNullifyVariant(CompilationContext $compilationContext)
      */
     public function isComplex()
     {
-        return isset($this->complexTypes[$this->type]);
+        return isset(self::complexTypes[$this->type]);
```
sergeyklay: `self::complexTypes` => `self::$complexTypes`

sergeyklay (changes requested):

PR #1533, opened 2017-05-08
Fix superglobals (2)

sergeyklay:


sergeyklay:
For Zephir code:

```zep
	public function remove(string index)
	{
		unset _SESSION[index];
	}
```

the following C-code was generated:
```c
PHP_METHOD(Test_Globals_Session_Base, remove) {

	zval *index_param = NULL, *_SESSION;
	zval *index = NULL;

	ZEPHIR_MM_GROW();
	zephir_fetch_params(1, 1, 0, &index_param);

	zephir_get_strval(index, index_param);


	zephir_array_unset(&_SESSION, index, PH_SEPARATE);
	ZEPHIR_MM_RESTORE();

}
```

Well, it should be:
```diff
PHP_METHOD(Test_Globals_Session_Base, remove) {

	zval *index_param = NULL, *_SESSION;
	zval *index = NULL;

	ZEPHIR_MM_GROW();
++	zephir_get_global(&_SESSION, SS("_SESSION") TSRMLS_CC);
	zephir_fetch_params(1, 1, 0, &index_param);

	zephir_get_strval(index, index_param);


	zephir_array_unset(&_SESSION, index, PH_SEPARATE);
	ZEPHIR_MM_RESTORE();

}
```


Cc: @sjinks

Library/Variable.php
```diff
@@ -1025,12 +1025,7 @@ public function isComplex()
      */
     public function isSuperGlobal()
     {
-        if (!$this->isExternal) {
-            return false;
-        }
-
-        $name = $this->getName();
-        return isset(self::$superglobals[$name]);
+        return isset($this->superglobals[$this->name]);
```
sergeyklay: @sjinks We should'n use varables like `_POST` or `_SESSION` as internal variables. They always  should be external.

Library/Variable.php
```diff
@@ -152,29 +152,29 @@ class Variable
...
-    );
+    ];
 
     /**
      * Names of superglobal variables
      *
      * @var array
      */
-    protected static $superglobals = array(
+    protected $superglobals = [
```
sergeyklay, line 169: @sjinks Do not use static variables/methods as log as it possible

PR #1535, opened 2017-05-10
Test static null comparsion

Jurigag:


Jurigag:
Don't merge it, trying to reproduce issue

Jurigag:
@sergeyklay i just tried to reproduce it to be similar to `Phalcon\Mvc\Collection::getReservedAttributes()`, now this PR should trigger problem.

Jurigag:
@sjinks 

There is some weird issue that when extend zephir class in php then something weird happens to static properties, like value of them is lost or they are even lost at all.

When i added var_dump it returns `&NULL`

PR #1550, opened 2017-06-20
drop zend_error_noreturn usage

remicollet:
See https://github.com/phalcon/cphalcon/issues/12909

Notice: **zend_error_noreturn** does exactly the same than **zend_error**, only exists to avoid compiler warnings

`
void zend_error_noreturn(int type, const char *format, ...) __attribute__ ((alias("zend_error"),noreturn));`


sergeyklay:
Thank you!

remicollet:
Notice: a new release will be very welcome, so we can build phalcon "tagged" version using "zephir" tagged version

sergeyklay:
I'll do it today

sergeyklay:
@remicollet done https://github.com/phalcon/zephir/releases/tag/0.9.9

remicollet:
Thanks

PR #1556, opened 2017-06-23
Test method from php class in zephir

Jurigag:


sergeyklay (changes requested):
Class 'Test\Mcall\McallAbstract' does not implement method: 'testExtending'

Jurigag:
Oh right, i thought that in phalcon dispatcher there is just somewhere controller class, there is not, let me update.

Issue #1530, opened 2017-05-02
Access to undeclared static property


xqkeji:
When ajax request more than 3 times, there will be "Access to undeclared static property"
E.g
```zep
namespace XqKeji;

class Di
{
	private static _default;

	public static function getInstance()
	{
		var di
		let di = self::_default;

		if !di {
			let di = new self();
			let self::_default = di;
		}

		return self::_default;
	}
}
```

```zep
class app
{
	public function __construct()
	{
		var di;
		let di = \XqKeji\Di::getInstance();
	}
}
```
Test:

```php
<?php
$App = new App();

```
 in window php7.0 ts

xqkeji:
`zephir_read_static_property_ce(&di, xqkeji_di_ce, SL("_default"), PH_NOISY_CC);`

Does php7 do not need TSRMLS_CC

xqkeji:
As long as the load is high, will appear "Access to undeclared static property"。

Jurigag:
Add that this happens only on php apache module, on cli or fpm it works fine and without issues.

sergeyklay:
Refs:
* https://github.com/phalcon/cphalcon/issues/12056
* https://github.com/phalcon/cphalcon/issues/12803
* https://github.com/phalcon/cphalcon/issues/12791
* https://github.com/phalcon/cphalcon/issues/12110
* https://github.com/phalcon/cphalcon/issues/12065
* https://github.com/phalcon/cphalcon/issues/12329

sergeyklay:
@sjinks Could you please take a look

hakimio:
@sjinks It's rather important that you take a look at this bug as soon as possible, because it prevents usage of thread-safe extension on any platform, not only windows. So, it's no longer possible to use phalcon with pthreads and similar stuff which actually requires thread safety.

afbora:
I got this error on Phalcon 3.2.0 today.
Its gone when i refresh the page 😕

sergeyklay:
@php318 Could you change your code to:
```diff
namespace XqKeji;

class Di
{
	private static _default;

	public static function getInstance()
	{
-		var di
-		let di = self::_default;

-		if !di {
+		if typeof self::_default !== "object" {
-			let di = new self();
-			let self::_default = di;
+			let self::_default = new self();
		}

		return self::_default;
	}
}
```

and test it

sergeyklay:
/ cc: @hakimio, @afbora

Jurigag:
[...]

PR #1592, opened 2017-10-28
Deal with #1530

sergeyklay:


Jurigag:
Well tests are passing, you need to test it on TS php i think.

Failed travis are because of valgrind, not sure if related or not.

sergeyklay:
In fact, I'm aware of the problem root. This PR just debug, not solution. And I have all of required ;)
```
$ ls -1 $(phpenv root)/versions/
5.5.38-nts-debug
5.6.31-nts-debug
5.6.31-zts-debug
7.0.25-nts-debug
7.0.25-zts-debug
7.1.11-nts-debug
7.1.11-zts-debug
system
```

PR #1616, opened 2017-12-23
fix global function has not parameter, compile failed

chrysanthemum:
zephir compile:
Error: Call to a member function getParameters() on null in /../zephir/Library/Compiler.php on line 2019


chrysanthemum:
like this in "fcall.zep" file
```zep
function fun1()  //  has not parameter
{
	echo "someting";
}

class Fcall
{
... 
```

sergeyklay:
@chrysanthemum This PR has conflicts that should be resolved. Could you please rebase?

chilimatic:
shouldn't it be 
```
if ($parameters !== null && count($parameters)) {
```
otherwise you could just use 
```
if ($parameters) {
```
would be less opcode operations and would have the complete same result.

chrysanthemum:
@sergeyklay thinks for your help!  is it correct now?

chrysanthemum:
@chilimatic I was almost mistaken by you. $parameters->getParameters() Be sure to execute this method.

chilimatic:
@chrysanthemum sorry I assumed an overloaded array. I will make sure to read up next time :)

sergeyklay:
Thank you

PR #1654, opened 2018-06-03
Fix some testing settings

danhunsaker:
Make tests more portable, so they can be run on users' systems as easily as on CI services.

- Load `test` module dynamically if it isn't globally installed.
- Use `composer.json` to indicate that PHPUnit 5.7 is acceptable when 4.8 isn't required, and allow Composer to determine which is appropriate.
- Fix `Issue1404Test.php` to use a regex that allows the version string to include distro data, and silently ignore it.

Also improve the way Travis tests are run:

- Run tests without Valgrind to get detailed output and code coverage without risking timeouts, then with Valgrind but no code coverage or detailed output to get memory leak information as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,11 @@ dist: trusty
 sudo: false
 
 php:
-  - 5.5
-  - 5.6
-  - 7.0
-  - 7.1
-  - 7.2
+  - '5.5'
+  - '5.6'
+  - '7.0'
+  - '7.1'
+  - '7.2'
 
 git:
   depth: 1
@@ -51,13 +51,13 @@ matrix:
       php: nightly
       compiler: clang
     - env: CPPFLAGS=-DZEPHIR_RELEASE CC=gcc
-      php: 7.1
+      php: '7.1'
     - env: CPPFLAGS=-DZEPHIR_RELEASE CC=clang
-      php: 7.1
+      php: '7.1'
     - env: CPPFLAGS=-DZEPHIR_RELEASE CC=gcc
-      php: 7.2
+      php: '7.2'
     - env: CPPFLAGS=-DZEPHIR_RELEASE CC=clang
-      php: 7.2
+      php: '7.2'
 
 cache:
   apt: true
@@ -77,11 +77,6 @@ before_install:
 
 install:
   - composer install --prefer-source --no-suggest
-  - |
-        if [ "${PHP_MAJOR}.${PHP_MINOR}" != "5.5" ]; then
-            composer remove --dev phpunit/phpunit
-            composer require -q -n --dev --no-progress --prefer-dist --no-suggest "phpunit/phpunit:5.7.*"
-        fi
   - bash ./unit-tests/ci/install-re2c $RE2C_VERSION
   - bash ./unit-tests/ci/install_zephir_parser.sh
   - ./install
@@ -91,7 +86,15 @@ before_script:
   - $(phpenv which php) compiler.php generate -Wnonexistent-function -Wnonexistent-class -Wunused-variable
   - $(phpenv which php) compiler.php stubs
   - $(phpenv which php) compiler.php api
-  - (cd ext; $(phpenv which phpize) && ./configure --silent --with-php-config=$(phpenv which php-config) --enable-test && make -j"$(getconf _NPROCESSORS_ONLN)" && make --silent install && phpenv config-add ../unit-tests/ci/test.ini)
+  - |
+    (
+      set -e
+      cd ext
+      $(phpenv which phpize)
+      ./configure --silent --with-php-config=$(phpenv which php-config) --enable-test
+      make -j"$(getconf _NPROCESSORS_ONLN)"
+      make --silent install
+    )
   #- ls -1 `$(phpenv which php-config) --extension-dir`
   #- phpenv versions
   - ulimit -c unlimited || true
@@ -100,9 +103,15 @@ before_script:
   #- sudo chmod +s $(which gdb)
 
 script:
-  - echo 'variables_order=EGPCS' >> "$(phpenv root)/versions/$(phpenv version-name)/etc/php.ini"
+  - echo -e 'variables_order=EGPCS\nenable_dl=true' >> "$(phpenv root)/versions/$(phpenv version-name)/etc/php.ini"
   - vendor/bin/phpcs --standard=PSR2 --report=emacs --extensions=php --warning-severity=0 Library/ unit-tests/Extension/ unit-tests/Zephir/
-  - |
+  - | # Tests and coverage
+      ./unit-tests/phpunit \
+        --not-exit \
+        -c phpunit.xml.dist \
+        --debug \
+        unit-tests/
+  - | # Check for memory leaks
       valgrind \
         --read-var-info=yes \
         --error-exitcode=1 \
@@ -114,7 +123,7 @@ script:
         ./unit-tests/phpunit \
           --not-exit \
           -c phpunit.xml.dist \
-          --debug \
+          --no-coverage \
           unit-tests/
   - $(phpenv which php) unit-tests/microbench.php
 
--- a/unit-tests/Extension/ExitDieTest.php
+++ b/unit-tests/Extension/ExitDieTest.php
@@ -23,6 +23,8 @@ class ExitDieTest extends \PHPUnit_Framework_TestCase
             $phpBinary .= ' -qrr';
         }
 
+        $phpBinary .= " -d 'enable_dl=true'";
+
         $testfile1 = __DIR__ .'/fixtures/exit.php';
         $return1 = `$phpBinary $testfile1`;
         $this->assertSame('', trim($return1));
--- a/.editorconfig
+++ b/.editorconfig
@@ -42,3 +42,7 @@ end_of_line              = lf
 indent_style             = space
 trim_trailing_whitespace = false
 insert_final_newline     = true
+
+[*.yml]
+indent_size              = 2
+indent_style             = space
--- a/Library/Optimizers/FunctionCall/IsPhpVersionOptimizer.php
+++ b/Library/Optimizers/FunctionCall/IsPhpVersionOptimizer.php
@@ -59,7 +59,7 @@ class IsPhpVersionOptimizer extends OptimizerAbstract
             throw new CompilerException("This function requires a scalar types parameter, $variableType given", $expression);
         }
 
-        preg_match('/^(?<major>\d+)(?:\.(?<minor>!?\d+))?(?:\.(?<patch>!?\d+))?$/', $expression['parameters'][0]['parameter']['value'], $matches);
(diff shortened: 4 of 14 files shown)


AlexNDRmac:
Regexp modification in Issue1404 also must be added into Zephir Lang Core. @sergeyklay - if it better, I`ll add modified Regexp into Core before this pull.

danhunsaker:
~~Actually, no, I looked at what Zephir's doing with version number checks, and it doesn't actually use a regex in the first place. Otherwise I'd've changed that, already, too.~~

Somehow missed that the optimizer had some of the version checking logic in it. Applied the regex to it, as well.

unit-tests/Extension/ExitDieTest.php
```diff
@@ -23,6 +23,8 @@ public function testExitDie()
             $phpBinary .= ' -qrr';
         }
 
+        $phpBinary .= " -d 'enable_dl=true'";
+
```
sergeyklay, line 27: @danhunsaker This deprecated feature will certainly be removed in the future: http://php.net/manual/en/info.configuration.php Why do we need this?

.travis.yml
```diff
@@ -100,9 +109,15 @@ before_script:
   #- sudo chmod +s $(which gdb)
 
 script:
-  - echo 'variables_order=EGPCS' >> "$(phpenv root)/versions/$(phpenv version-name)/etc/php.ini"
+  - echo -e 'variables_order=EGPCS\nenable_dl=true' >> "$(phpenv root)/versions/$(phpenv version-name)/etc/php.ini"
```
sergeyklay, line 106: Could you please create an ini file like:
```ini
variables_order=EGPCS
```
and then add this file via:
```yaml
- phpenv config-add unit-tests/ci/zephir.ini
```

This approach will allow us to add more settings in future to this file without using echo. Also as you can see the **`enable_dl`** is deprecated feature and will be removed in the future: http://php.net/manual/en/info.configuration.php Why do we need this?

.travis.yml
```diff
@@ -77,11 +77,11 @@ before_install:
...
-  - |
-        if [ "${PHP_MAJOR}.${PHP_MINOR}" != "5.5" ]; then
-            composer remove --dev phpunit/phpunit
-            composer require -q -n --dev --no-progress --prefer-dist --no-suggest "phpunit/phpunit:5.7.*"
-        fi
+  # - |
+  #       if [ "${PHP_MAJOR}.${PHP_MINOR}" != "5.5" ]; then
+  #           composer remove --dev phpunit/phpunit
+  #           composer require -q -n --dev --no-progress --prefer-dist --no-suggest "phpunit/phpunit:5.7.*"
+  #       fi
```
sergeyklay: This change led to problem:

```
PHP Strict standards:  Declaration of Extension\Issue1404Test::onNotSuccessfulTest() should be compatible with PHPUnit_Framework_TestCase::onNotSuccessfulTest(Exception $e) in /home/travis/build/phalcon/zephir/unit-tests/Extension/Issue1404Test.php on line 250
PHP Stack trace:
PHP   1. {main}() /home/travis/build/phalcon/zephir/unit-tests/phpunit:0
PHP   2. PHPUnit_TextUI_Command::main() /home/travis/build/phalcon/zephir/unit-tests/phpunit:38
PHP   3. PHPUnit_TextUI_Command->run() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/TextUI/Command.php:100
PHP   4. PHPUnit_Runner_BaseTestRunner->getTest() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/TextUI/Command.php:123
PHP   5. PHPUnit_Framework_TestSuite->addTestFiles() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/Runner/BaseTestRunner.php:58
PHP   6. PHPUnit_Framework_TestSuite->addTestFile() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/Framework/TestSuite.php:384
PHP   7. PHPUnit_Util_Fileloader::checkAndLoad() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/Framework/TestSuite.php:310
PHP   8. PHPUnit_Util_Fileloader::load() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/Util/Fileloader.php:38
Strict standards: Declaration of Extension\Issue1404Test::onNotSuccessfulTest() should be compatible with PHPUnit_Framework_TestCase::onNotSuccessfulTest(Exception $e) in /home/travis/build/phalcon/zephir/unit-tests/Extension/Issue1404Test.php on line 250
Call Stack:
    0.0002          0   1. {main}() /home/travis/build/phalcon/zephir/unit-tests/phpunit:0
    0.0136          0   2. PHPUnit_TextUI_Command::main() /home/travis/build/phalcon/zephir/unit-tests/phpunit:38
    0.0137          0   3. PHPUnit_TextUI_Command->run() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/TextUI/Command.php:100
    0.2155          0   4. PHPUnit_Runner_BaseTestRunner->getTest() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/TextUI/Command.php:123
    0.2470          0   5. PHPUnit_Framework_TestSuite->addTestFiles() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/Runner/BaseTestRunner.php:58
    0.7182          0   6. PHPUnit_Framework_TestSuite->addTestFile() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/Framework/TestSuite.php:384
    0.7187          0   7. PHPUnit_Util_Fileloader::checkAndLoad() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/Framework/TestSuite.php:310
    0.7188          0   8. PHPUnit_Util_Fileloader::load() /home/travis/build/phalcon/zephir/vendor/phpunit/phpunit/src/Util/Fileloader.php:38
```

Please take a look at https://travis-ci.org/phalcon/zephir/jobs/387499801

.travis.yml
```diff
@@ -5,11 +5,11 @@ dist: trusty
...
-  - 5.5
-  - 5.6
-  - 7.0
-  - 7.1
-  - 7.2
+  - '5.5'
+  - '5.6'
+  - '7.0'
+  - '7.1'
+  - '7.2'
```
sergeyklay, line 12: Actually I don't know why do we need these changes. As I know YAML works fine with string w/o spaces and they do not need escaping. Could you elaborate on this?

sergeyklay (changes requested):

unit-tests/Extension/ExitDieTest.php
danhunsaker: Interesting that they say that. A change to `dl()` was just made in 7.2 to allow extensions to be loaded by extension name rather than filename. Hardly an expected behavior for a deprecated feature.

It's needed to be able to load the extension without using an ini file change. More on that below.

.travis.yml
danhunsaker: These are not treated as strings, without the quotes, and are in fact treated as floats. Type juggling allows this to sometimes work as expected, but I don't like to rely on it, given the way floats handle rounding and often even formatting (I've encountered errors where `2.2` becomes `'2.200000'`, for example, which of course doesn't match as expected for a version number, as it isn't a float at all). I recognize it's a preference thing, though, so I can revert that if we'd prefer to just leave it alone.

.travis.yml
danhunsaker: Ugh. Ok. PHPUnit itself is inconsistent on this one.

I changed it because I was seeing the opposite error on my system (and in Travis, for 5.6+ - see https://travis-ci.org/phalcon/zephir/jobs/387590228#L2107, for example), where the method in question does _not_ have the type hint, and PHP is complaining that the child class _does_. Since it's more correct for child classes to relax type restrictions than to be more strict about them (a change in 7.1 or 7.2, if I remember correctly, made this explicit by removing the error messages for that case), I think the revision is the more correct one, but we might be able to figure out another approach that lets the class support both parents, somehow...

.travis.yml
danhunsaker: The current (unpatched) approach already uses a separate ini file (though admittedly I ~~commented that part out~~ removed that part, above) and a change to the core ini (the line we're discussing here) to make things function as needed for these tests. This approach is very inflexible and non-portable.

What I mean, here, is that the current approach requires changes to the system-wide configuration of PHP itself, something which isn't always desirable, or in some cases even possible. This prevents running tests locally, meaning Travis becomes the only testing platform available, and every change will likely result in several test runs to resolve any problems. Alternately, users who contribute to Zephir dev will need to have the `test` extension loaded across their entire system, at least for the duration of tests, and are responsible for remembering to add and/or remove this entry every time manually.

The primary purpose of this PR is to remove any need to modify the ini files at all to successfully run tests. One option is to load the extension directly via the CLI, using `php -d 'extension=test' ./unit-test/phpunit ...`, but this approach actually leads to memory leaks that cause Valgrind to fail the tests. So I've opted for the alternative where the test bootstraps load the extension themselves, assuming it isn't loaded already, which works perfectly in every case, and allows tests to be run locally as well. (The PHPUnit docs say that it supports the `-d` option to its CLI command, but I wasn't able to get that working in any way, so here we are.)

.travis.yml
sergeyklay: @danhunsaker Can you please try to sort out with this?

danhunsaker: I just realized I was responding to the error message rather than the actual changes being flagged. My above comment applies to a change in the `Issue1404Test` class itself, wherein I dropped the type hint. That, ultimately, is the actual cause of the error message above.

What's going on _here_ is that I moved the "use PHPUnit 5.7 if we're not using PHP 5.5" logic to Composer's realm of responsibility, by adjusting the version requirement in `composer.json` to allow `4.8` _or_ `5.7`, as appropriate. Since 5.7 doesn't support PHP 5.5, but does support 5.6 and 7.x, Composer automatically installs 4.8 for PHP 5.5, and 5.7 for all the others. No shell commands required.

~~I probably ought to remove those lines rather than comment them out, though.~~ Removed in most recent amendment.

danhunsaker: As to the inheritance error (which we now only get in 5.5, instead of 5.6, 7.0, 7.1, and 7.2), I'll see what I can figure out.

danhunsaker: Hmmm. None of the approaches I can think of to conditionally define this method are working. I tried:
- dropping it into a trait, and then defining that trait based on the PHP version; PHPCS complained about putting the trait into the same file as the class
- dropping the trait into a separate file; PHPCS complained that the keyword `trait` appeared twice, and only one trait/class/interface should be defined per file (which is fair enough)
- a trait file for the autoloader that `require`s the correct _actual_ trait file depending on PHP version; the `require` statements seem to not be executed, so the trait is never actually defined before it gets used
- attempting to override the method in `__construct`; just doesn't do anything

I might be able to get away with manually `require`ing the appropriate trait file _manually_ instead of relying on the autoloader.

danhunsaker: OK, that worked. This message should be gone.

sergeyklay (approved):

sergeyklay:
Thank you!

PR #1655, opened 2018-06-09
Allow extension to be loaded prior to the tests

danhunsaker:
If `extension=test` (or `extension=php_test.dll`) is in the `php.ini`, or otherwise loaded prior to the tests themselves, the tests should be satisfied with that and continue on running. Unfortunately, I flubbed the logic there, and made it so the extension _had_ to be loaded by the tests themselves, or not at all.

So, uh, let's fix that!


AlexNDRmac (approved):

sergeyklay (approved):

sergeyklay:
Thank you

PR #1658, opened 2018-06-29
Add type hints for scalar arguments and return values in Zend3

### LINKED ISSUE #1656: Compile Type Hints?
PHP has supported type hinting function arguments for some time, now, and Zend Engine has supported the same hints in extensions for just as long. Zephir has implemented support for the non-scalar types supported in PHP 5.x, and these work fine. But it currently _doesn't_ export _scalar_ or _return_ type hints (supported since PHP 7.0) at all, so userspace classes that extend or implement extension classes or interfaces need to do so without the appropriate type hinting one would expect to see, there.

I've already started work on adding these type hints to the compiled output, and it appears to be working as expected. That said, there's one point in particular where things get tricky, and that's with _strict_ typing of arguments. PHP will automatically use argument type hints to coerce values to the expected type, from whatever type they happen to be, according to its normal type conversion rules/logic. That means every value passed in to a Zephir function/method will always pass the strictness tests we currently compile, and we'll therefore never be able to catch attempts to pass in the wrong type - even if PHP can't coerce the value itself, it'll fail before we ever see it. Currently the only way to disable this coercion is to `declare(strict_types=1)` in every file that _calls into_ a Zephir extension, and even then PHP handles rejecting the values before Zephir even sees them.

The purpose of this issue (#1656) is twofold. One, I'm wondering how users would expect this feature to work when building extensions with type hinted function/method arguments. And two, I'm reaching out to see if anyone else can think of a way to work around the automatic coercion of PHP to restore the current strict argument typing support, or possibly redesign the strict argument typing feature to work around PHP.

Is it even desired to export these type hints to Zend and PHP? If so, would it be acceptable to omit them when strict typing is desired (that is, not tell PHP what the types are expected to be, so Zephir can reject them rather than allow them to be converted)? Assuming we want such hints, but aren't willing to leave them out for strict types, we'll need to figure out how to get PHP to tell us what type was originally passed to us, if not how to make it simply pass the original value itself.

I've mentioned this a couple of times in the Phalcon Discord server's `#zephir` channel, but it doesn't seem to see much activity, so I thought I'd post something more detailed here for review and discussion.

danhunsaker:
Fixes #1656


danhunsaker:
Discussion of the failed checks, and how to handle them properly, was started in #1656 before this WIP was submitted for reference.

sergeyklay:
There were 6 failures:

1) Extension\MCallTest::testOptionalParameterBooleanException
Failed asserting that exception of type "\InvalidArgumentException" is thrown.

2) Extension\Oo\OoParamsTest::testSetStrictAgeException1
Failed asserting that exception of type "\InvalidArgumentException" is thrown.

3) Extension\Oo\OoParamsTest::testSetStrictAgeException2
Failed asserting that exception of type "\InvalidArgumentException" is thrown.

4) Extension\Oo\OoParamsTest::testSetStrictAverageException1
Failed asserting that exception of type "\InvalidArgumentException" is thrown.

5) Extension\Oo\OoParamsTest::testSetStrictAverageException2
Failed asserting that exception of type "\InvalidArgumentException" is thrown.

6) Extension\Oo\OoParamsTest::testSetStrictNameException
Failed asserting that exception of type "\InvalidArgumentException" is thrown.

unit-tests/Zephir/Test/TypeHintsTest.php
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        if (file_exists('ext')) {
+            exec('rm -rf ext/');
+        }
+
+        if (getcwd() != $this->pwd) {
+            chdir($this->pwd);
+        }
+    }
```
sergeyklay: Could you adopt this to use in AppVeyor (Windows) too

unit-tests/Zephir/Test/TypeHintsTest.php
```diff
@@ -0,0 +1,114 @@
...
+     * Restore current directory, and clean config.json.
+     */
+    public function tearDown()
+    {
+        if (file_exists('.temp')) {
+            exec('rm -rf .temp/');
+        }
+
+        if (file_exists('ext')) {
+            exec('rm -rf ext/');
```
sergeyklay: Could you adopt this to use in AppVeyor (Windows) too

sergeyklay (changes requested):

unit-tests/Zephir/Test/TypeHintsTest.php
danhunsaker: Hmm. fair point. I'll need to take a peek at the other test I pulled this block from and adjust it, too.

danhunsaker: This is addressed. Now handling the other issues that are preventing this WIP from becoming a true PR.

danhunsaker:
OK, should be ready to go. The most recent change was to start using the PHP version to detect whether we have support for strict typing or not, and adjusting the expected exception accordingly. With `declare(strict_types=1)` in the files where strict type checks are appropriate, PHP throws `TypeError` before we see values passed to Zephir code. Otherwise, PHP coerces values to the expected type automatically, also before we see them. PHP 5.x doesn't support strict typing in any of its incarnations, nor does it support type hinting in the first place, so Zephir code will compile to the same C code as before, with ZE2.

sergeyklay (approved):

sergeyklay (approved):

sergeyklay:
Thank you!

PR #1684, opened 2018-10-21
Support array return type

danhunsaker:
Missed a few spots where arrays needed to be handled, but weren't already.


sergeyklay:
Fixed. Thank you

sergeyklay:
Fixed. Thank you

PR #1717, opened 2018-11-13
Provide Windows entrypoint (zephir.bat)

sergeyklay:


danhunsaker (commented):
.

zephir.bat
```diff
@@ -0,0 +1,18 @@
...
+cls
+if "%PHPBIN%" == "" set PHPBIN=@php_bin@
+if exist "zephir" goto INTERNAL
+if not exist "%PHPBIN%" if "%PHP_PEAR_PHP_BIN%" neq "" goto USE_PEAR_PATH
+GOTO RUN
+:USE_PEAR_PATH
+set PHPBIN=%PHP_PEAR_PHP_BIN%
+:RUN
+"%PHPBIN%" "@bin_dir@\zephir" %*
+:INTERNAL
```
danhunsaker: Probably want to step over `:INTERNAL` if `:RUN` was reached?

sergeyklay: @danhunsaker I simplified the .bat file. It's using `%PHP_PEAR_PHP_BIN%` because that's a global variable which may be set by PEAR. I just figure, if it's available anyway, why not use it; and in the other case, it will just try php, assuming most Windows users will have added that to their `PATH`.

danhunsaker: That's not an issue; the logic itself makes sense. The thing I'm concerned about is that once it hits `:RUN`, and runs the command, it _then_ continues on into `:INTERNAL`, where it will attempt to run the command again using the internal approach, which will either fail with an error message, or succeed and the command will be run twice. We probably want to either `goto` a label at the end of the file, or end the script execution entirely, before the batch file hits `:INTERNAL` and its commands.

Issue #1713, opened 2018-11-11
0.10.12 broke "require" call while running in a phar archive and other concerns about "require" changing for a future Phalcon 3 LTS bugfix version.


dschissler:
@sergeyklay 

My VM is a standard Ubuntu 18.04, PHP 7.2, latest zephir parser extension, latest Phalcon 3 LTS.

I see that 0.10.12 advertised as initial PHP 7.3 support.  Something changed between 0.10.11 and 0.10.12 that breaks the "phar://" protocol from working correctly when called from zephir code that is run in [initiated from] a phar.  I've tested it twice now after switching back and forth, deleting the extension cache directories each time.

I know that the behavior of Zephir `require` has never worked the same/correctly in PHP 7 as it did in PHP 5, but whatever you did to to improve this has now broken phar use within zephir.

I'm concerned that also the behavior of the `require` has changed for the stable branch of Phalcon 3, because it never worked correctly and if you have now fixed it in Zephir then there could be different behavior for someone's Phalcon 3 setup (even if now possibly correct and broken before).  I see that a new Phalcon 3 has not yet been released, so that is a good thing IMO.

I encourage you to set aside the `0.10.11` zephir branch as **possibly** (not certainly) the zephir version that is most compatible with how things have been done.  I don't believe that it was correct to have made whatever changes you made on a "bugfix" version of zephir but to have instead created the 0.11.0.

I believe that considering how broken the `require` has been for the entire life of Phalcon 3 that even fixing it at the very end of its support is the incorrect thing to do, since it changes behavior.

dschissler:
For now I'll beusing the latest Zephir and ignoring this one broken aspect for now and when I need to be working with phars then I'll revert to the `0.10.11` version.

dschissler:
[...]

PR #1731, opened 2018-11-16
Fixed "require" regression introduced in the 0.10.12

sergeyklay:
Closes: #1713


kernels/ZendEngine3/require.c
```diff
@@ -42,17 +33,8 @@ int zephir_require_ret(zval *return_value_ptr, const char *require_path)
 	zend_file_handle file_handle;
 	zend_op_array *new_op_array;
 	zval dummy, local_retval;
-	char realpath[MAXPATHLEN];
 	int ret;
 
-	if (UNEXPECTED(!VCWD_REALPATH(require_path, realpath))) {
```
dschissler, line 48: of course.  Thanks.

PR #1806, opened 2019-02-17
Fix #1660

### LINKED ISSUE #1660: Segmentation fault when using $return_value
If declaring a function that returns bool with a return variable named $return_value the generated extension crashes.

```php
namespace Test;
class SomeTest
{
    public static function bug() -> bool
    {
        bool $return_value = true;
        return $return_value; /*Segfault*/
    }
}
```
It seems the generated code conflicts with the return_value variable declared by the zend engine for all functions. Took me a long debugging session to figure this out...

dreamsxin:


sergeyklay:
@dreamsxin Could you please send this patch to the `development` branch?

danhunsaker:
Not sure this actually addresses the issue in #1660... The variable's type doesn't need to be overridden; its name does. This code should still have the same issue originally reported, or at least a similar one - that we're dealing with a variable whose name is already in use within the code we generate, conflicting with the variable of the same name in the developer's code. Replacing the actual return type with `variable` masks the actual itself and prevents type hinting from working properly.

dreamsxin:
@danhunsaker In the future, if we allow kernel variables to be used directly and call kernel functions in Zephir, it would not be a good idea to modify variable names.

danhunsaker:
I'm not aware of any reason we'd want to allow access to the underlying Zephir kernel C constructs directly, nor what benefit that would provide. In any case, changing the variable's type is not the correct answer, as it violates the type strictness guarantees. Transparently changing the variable's C representation (say, with a prefix) to prevent collisions with Zephir-defined values is safe _and_ correct.

dreamsxin:
It is possible to prefix all variables instead of modifying return_value individually. @sergeyklay

PR #1807, opened 2019-02-17
Fix #1604

### LINKED ISSUE #1604: Duplicate extension key prefix in the global configuration
Here is the config.json
```js
{
    "namespace": "utils",
    "name": "utils",
    "description": "An utils helper extension",
    "author": "aiddroid",
    "version": "0.0.1",
    "verbose": false,
    "requires": {
        "extensions": []
    },
    "globals": {
        "utils.enable_cache": {
            "type": "bool",
            "default": true
        }
    }
}
```
but got "utils.utils.enable_cache" in phpinfo().
phpinfo

and after I removed the "utils" prefix from globals section in config.json, The key became "enable_cache" in phpinfo().It's very strange.
image

dreamsxin:


sergeyklay:
@dreamsxin Could you please send this patch to the `development` branch?

Library/Compiler.php
```diff
@@ -1340,7 +1340,7 @@ public function processExtensionGlobals($namespace)
                 }
 
                 if (!isset($iniEntry['name'])) {
-                    $iniName = $name;
+                    $iniName = $namespace.'.'.$name;
```
AlexNDRmac, line 1343: This is not fixes #1604. Because its not about how ini Entry pass to `Compiler`. Issue #1604 - more wider then just fix PHP code as Compiler pre-processes. This issue requires some changes in C-code generation. So, if you will test your code changes, you can see that duplicates of namespaces are still present in `phpinfo()` as in CLI output and as in HTML output.

dreamsxin: The reason for the #1604 duplication is that namespace and prefix use the same word. The correct treatment is to add namespace.

Contain dot  will add namespace in `Code\Builder\Struct::getInitEntry`.

dreamsxin:
@sergeyklay @AlexNDRmac 
We can discuss the rules for adding prefixes, about containing or without points.
1. `xx.xx`
```php
$namespace.'.'.$name;
```
2. `xx`
```php
$namespace.'.'.$name;
```

Library/Compiler.php
AlexNDRmac: As I said... for this issue we need change C-code generation also. I tested fix such as you provided and saw that some variables aren't show in `phpinfo()`.

For example: `/ext/test.c`
```c
PHP_INI_BEGIN()
    STD_PHP_INI_BOOLEAN("test.test.my_setting_1", "0", PHP_INI_ALL, OnUpdateBool, test.my_setting_1, zend_test_globals, test_globals)
    ...
PHP_INI_END()
```

sergeyklay:
@dreamsxin As I know @AlexNDRmac already done some work in this direction. Maybe he should do a PR on which we could work together.

Library/Compiler.php
dreamsxin: I don't understand what you mean, Ccustom prefix name,  should use `ini-entry`, Otherwise, namespace is prefixed.

dreamsxin:
I close that.

PR #1810, opened 2019-02-19
 Move cache and logs to the common place 

sergeyklay:
Moved `cache` and `logs` directories to the user's home. 
  - On macOS Zephir will use `XDG` if it is possible, otherwise `$HOME/Library`
  - On Windows Zephir will use `LOCALAPPDATA` if it is possible, otherwise home dir as a base path
  - In any other cases, e.g. Linux, BSD and so on Zephir will use `XDG`if it is possible,
    otherwise `$HOME/.local` and `$HOME/.cache`


CODE_OF_CONDUCT.md
```diff
@@ -1,76 +1,25 @@
 # Contributor Covenant Code of Conduct
 
-## Our Pledge
+## TL;DR
```
danhunsaker: Why are we changing the CoC? Especially in this PR...

Library/functions.php
```diff
@@ -206,7 +207,7 @@ function is_windows()
  */
 function is_macos()
 {
-    return 'DARWIN' === strtoupper(substr(PHP_OS, 0, 6));
+    return 0 === stripos(PHP_OS, 'DARWIN');
 }
```
danhunsaker, line 212: `strtolower()` is extraneous with a case-insensitive check.

danhunsaker (changes requested):
Minor point and major question. Otherwise looks good.

CODE_OF_CONDUCT.md
sergeyklay: My bad I'll revert

Library/functions.php
sergeyklay: @danhunsaker  `stripos`  the same as `strtoupper` (`strtolower`) + `substr`, isn't

danhunsaker: This stupid thing put the comment in the wrong part of the code. This specific line is perfectly fine (though I'm of a mind to use `-1 <=` instead of `0 ===`).

Library/functions.php
```diff
@@ -218,7 +219,7 @@ function is_macos()
  */
 function is_bsd()
 {
-    return false !== stristr(strtolower(PHP_OS), 'bsd');
+    return false !== stripos(strtolower(PHP_OS), 'bsd');
```
danhunsaker: _This_ is the change I meant when I said `strtolower()` is extraneous.

PR #1811, opened 2019-02-20
Fix #1751 add context in class CompilerFileAnonymous

### LINKED ISSUE #1751: Anonymous function
```zep

namespace Work\Abc;

use Work\Task\Timer;

class Test {


    public function a()
    {
        var result;

	Timer::start(function () {

            let result = Timer::wait();

         });	
     }
}
```
[ERROR] Class name: Work\Timer does not exist in `let result = Timer::wait();`

dreamsxin:
Pass context to Closure's compiler


dreamsxin:
Pass context to Closure's compiler

PR #1813, opened 2019-02-20
Fix #1775

### LINKED ISSUE #1775: Segmentation Fault: Manipulating an undefined variable as an array
When running PHP as CLI `$_SESSION` is undefined. So attempting to call this function will attempt to append an array element to an undefined variable which results in a segmentation fault.

```zep
function set(string index, var value)
{
    var uniqueId;

    let uniqueId = this->_uniqueId;
    if !empty uniqueId {
        let _SESSION[uniqueId . "#" . index] = value;
        return;
    }

    let _SESSION[index] = value;
}

set("test"); // segfault
```

Thanks :)

dreamsxin:
This PR generate tmp array and add to `&EG(symbol_table)`.

### DIFF
--- a/Library/Backends/ZendEngine3/Backend.php
+++ b/Library/Backends/ZendEngine3/Backend.php
@@ -61,7 +61,6 @@ class Backend extends BackendZendEngine2
     public function getVariableCode(Variable $variable)
     {
         if ($variable->isDoublePointer() ||
-            $variable->isSuperGlobal() ||
             \in_array($variable->getName(), ['this_ptr', 'return_value']) ||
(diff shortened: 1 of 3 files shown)


dreamsxin:
@sergeyklay It's ok.

sergeyklay:
@dreamsxin Segmentation fault (core dumped) on Travis for all PHP 7.x versions. Did you see

dreamsxin:
@sergeyklay Check

sergeyklay (approved):

sergeyklay (approved):

sergeyklay:
Thank you

Issue #1828, opened 2019-03-10
Compilation error for instanceof


sergeyklay:
Steps to reproduce:

```zep
public function test07(string test) -> bool
{
    var obj;
    let obj = new \stdClass;

    return obj instanceof test;
}
```

The error message:
```
In file included from /home/klay/.phpenv/versions/7.3.0-zts-debug/include/php/Zend/zend.h:27:0,
                 from /home/klay/.phpenv/versions/7.3.0-zts-debug/include/php/main/php.h:33,
                 from /home/klay/workspace/c/zephir-issue-1802/ext/bug/instance.zep.c:6:
/home/klay/workspace/c/zephir-issue-1802/ext/bug/instance.zep.c: In function ‘zim_Bug_Instance_test07’:
/home/klay/.phpenv/versions/7.3.0-zts-debug/include/php/Zend/zend_types.h:649:39: error: invalid type argument of unary ‘*’ (have ‘zval {aka struct _zval_struct}’)
 #define Z_STRVAL_P(zval_p)   Z_STRVAL(*(zval_p))
```

joeyhub:
This has always been a problem in php, you have to use is_a.

sergeyklay:
Fixed in the  #1829

PR #1829, opened 2019-03-10
Fixed compilation error for instanceof

sergeyklay:
* Type: bug fix
* Link to issue: #1828

### DIFF
--- a/test/instanceoff.zep
+++ b/test/instanceoff.zep
@@ -7,28 +7,28 @@ namespace Test;
 
 class Instanceoff
 {
-	public function testInstanceOf1()
+	public function testInstanceOf1() -> boll
 	{
 		var a;
 		let a = new \stdClass();
 		return a instanceof \stdClass;
 	}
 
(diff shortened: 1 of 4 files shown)


test/instanceoff.zep
```diff
@@ -37,7 +37,7 @@ class Instanceoff
 		return false;
 	}
 
-	public function testInstanceOf5(var a)
+	public function testInstanceOf5(var a) -> boll
```
Jeckerson:
```suggestion
	public function testInstanceOf5(var a) -> bool
```

test/instanceoff.zep
```diff
@@ -7,28 +7,28 @@ namespace Test;
 
 class Instanceoff
 {
-	public function testInstanceOf1()
+	public function testInstanceOf1() -> boll
```
Jeckerson:
```suggestion
	public function testInstanceOf1() -> bool
```

PR #1868, opened 2019-06-06
fix incorrect usage of zend_declare_class_constant_ex

dzuelke:
https://github.com/phalcon/cphalcon/issues/14160
https://bugs.php.net/bug.php?id=78121

Hello!

This fixes https://github.com/phalcon/cphalcon/issues/14160, author attribution is for @krakjoe in the commit.

I haven't touched the changelog yet as I'm not sure what the target version would be? Surely a bugfix release and not the next major version?

PR is editable.

PR #1869, opened 2019-06-06
fix phalcon/cphalcon#14160 incorrect usage of zend_declare_class_constant in 7.2+

krakjoe:
see phalcon/cphalcon#14160


krakjoe:
closing, covered by #1868

PR #1887, opened 2019-07-16
Fix #1885: Check return value before creating dynamic class.

### LINKED ISSUE #1885: NewInstanceOperator fails to check returned class before initiating object.
Hello,
There is a problem when using the new instance operator. I believe I finally tracked down the issue that required the fix https://github.com/phalcon/cphalcon/pull/14238 in cphalcon. I'm not familiar with the source code for the php interpreter or zephir, so please bear with me.

Currently, when using the `new {classname}()` operator in `zep` files, the following C code is generated.
```
zephir_fetch_safe_class(&_0$$3, &definition);
_1$$3 = zephir_fetch_class_str_ex(Z_STRVAL_P(&_0$$3), Z_STRLEN_P(&_0$$3), ZEND_FETCH_CLASS_AUTO);
object_init_ex(&handler, _1$$3);
if (zephir_has_constructor(&handler TSRMLS_CC)) {
    ZEPHIR_CALL_METHOD(NULL, &handler, "__construct", NULL, 0);
    zephir_check_call_status();
}
```

The problem is that the function `zephir_fetch_class_str_ex` will return `NULL` if the autoloaded class as a syntax error. This goes unchecked when passing it into [`object_init_ex`](https://github.com/php/php-src/blob/0e5ecc7a8fec52ec6eb67b150b5bc7e4e3fd79b1/Zend/zend_API.c#L1226) which expects the handler to be valid since there isn't any guard in place.

That's why using [`zephir_create_instance()`](https://github.com/phalcon/zephir/blob/f5f8f2aa93192dd8027bcfb3eee4a2ce8fb5f835/kernels/ZendEngine3/object.c#L1137) instead of `new {classname}()` doesn't cause a segfault.

~~Hopefully someone with more knowledge of Zephir can implement this fix. If not, I should be able to work on it in a few weeks.~~ I'm working on a patch for this right now.

davidofferman:
Hello!

* Type: bug fix
* Link to issue: #1885 

- [x] I updated the CHANGELOG

Small description of change:
Code to check the `$classEntry` value before calling the method `Zephir\BaseBackend::initObject()` inside of `Zephir\Operators\Other\NewInstanceOperator`.

Without this check, it's possible that NULL could be passed through and end up causing a segment fault. Happens when auto-loading a class with a syntax error.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
 - Added initial bash completion support (see zephir-autocomplete file)
+
+### Fixed
+- Fixed segfault when auto-loading class with syntax error
+  [#1885](https://github.com/phalcon/zephir/issues/1885)
 
 ## [0.12.0] - 2019-06-20
 ### Added
--- a/Library/Operators/Other/NewInstanceOperator.php
+++ b/Library/Operators/Other/NewInstanceOperator.php
@@ -135,6 +135,12 @@ class NewInstanceOperator extends BaseOperator
                     $classNameToFetch = 'Z_STRVAL_P('.$safeSymbol.'), Z_STRLEN_P('.$safeSymbol.')';
                     $zendClassEntry = $compilationContext->cacheManager->getClassEntryCache()->get($classNameToFetch, true, $compilationContext);
                     $classEntry = $zendClassEntry->getName();
+
+                    $compilationContext->codePrinter->output('if(!'.$classEntry.') {');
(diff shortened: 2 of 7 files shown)


davidofferman:
I'll try to think of a better unit-test so we can pass the travis-ci build. Worst case scenario I can just remove the `php` extension from the file.

Give me a few days to get around to this unless you have something in mind.
Thanks.

sergeyklay:
Anyway you can exclude any file/directory from code style testing using https://github.com/phalcon/zephir/blob/development/.php_cs.dist

davidofferman:
Thanks for that information. I'll keep that in mind for the future. 

The latest commit resolves the problem and now passes the automated tests and works as expected.

sergeyklay:
Thank you

PR #1900, opened 2019-08-04
Introduced initial ability to generate zend_module_deps

sergeyklay:
Hello!

* Type: new feature
* Link to issue: [phalcon/cphalcon#13702](https://github.com/phalcon/cphalcon/issues/13702), [phalcon/cphalcon#13794](https://github.com/phalcon/cphalcon/pull/13794)

- [x] I updated the CHANGELOG

Small description of change:

Introduced initial ability to generate `zend_module_deps` by using `requires` from project configuration (`config.json`).

To do so, add to your `config.json` configuration like this:
```json
{
    "requires": {
        "extensions": [
            "module1",
            "module2",
            "module3"
        ]
    }
}
```

With such a config section Zephir will generate C-code similar to follows:
```diff
zend_function_entry php_test1_functions[] = {
	ZEND_FE_END

};
+
+static const zend_module_dep php_test1_deps[] = {
+	ZEND_MOD_REQUIRED("module1")
+	ZEND_MOD_REQUIRED("module2")
+	ZEND_MOD_REQUIRED("module3")
+	ZEND_MOD_END
+};

zend_module_entry test1_module_entry = {
	STANDARD_MODULE_HEADER_EX,
	NULL,
-	NULL,
+	php_test1_deps,
	PHP_TEST1_EXTNAME,
	php_test1_functions,
	PHP_MINIT(test1),
#ifndef ZEPHIR_RELEASE
	PHP_MSHUTDOWN(test1),
#else
	NULL,
#endif
	PHP_RINIT(test1),
	PHP_RSHUTDOWN(test1),
	PHP_MINFO(test1),
	PHP_TEST1_VERSION,
	ZEND_MODULE_GLOBALS(test1),
	PHP_GINIT(test1),
	PHP_GSHUTDOWN(test1),
#ifdef ZEPHIR_POST_REQUEST
	PHP_PRSHUTDOWN(test1),
#else
	NULL,
#endif
	STANDARD_MODULE_PROPERTIES_EX
};
```

From documentation:

>  The system checks for the eventually attached `zend_module_deps` to the `zend_module_entry`. This structure basically allows a PHP extension to declare extensions it depends on, and dependencies have types :

> - A PHP extension can tell the system that it is not compatible with another already registered extension, so it will fail loading itself into the engine
> - A PHP extension can tell the system that it requires another PHP extension to be loaded before itself

Right now we're going to use the second option.

> Here, only **conflicts are checked**, and as there is no particular registration order, a PHP extension A could declare being in conflict with B, but B could be loaded after A, hence the check won't work and you'll probably get some trouble. When dealing with conflicts, it is better to sort the extensions in the php.ini file as a FIFO. **PHP extensions will be registered in the exact same order they've been declared in php.ini**.

Thanks


ruudboon:
Is it possible to add the extension version?

sergeyklay:
@ruudboon Yes, but I would prefer to refrain from this at this stage. I'll think about a better implementation a bit later.

danhunsaker:
Looks good. I would definitely prefer to support "conflicts with" values as well, and to be able to list version numbers in both sections. But as a starting point, this is still really helpful!

PR #1903, opened 2019-08-08
add static property reference test case

zhaoyanliang8:
This test case is used to test static property reference of class. Currently, the version of php7.2 and below has not passed the test.


sergeyklay:
@zhaoyanliang2 Thank you for the bug report. We'll take a look ASAP.

sergeyklay:
Fixed in the `development` branch. Feel free to open a new issue if the problem appears again. Thank you for the bug report.

PR #2001, opened 2019-11-04
Fix #2000

### LINKED ISSUE #2000: Calling object methods from static context yields segmentation fault when internal-call-transformation: true
Hi fellas,

As recently as 0.12.11, the following block of code results in a segmentation fault ...

```zephir
    /**
     * Return the elementwise maximum of two vectors.
     *
     * @param \Tensor\Vector a
     * @param \Tensor\Vector b
     * @throws \InvalidArgumentException
     * @return self
     */
    public static function maximum(const <Vector> a, const <Vector> b) -> <Vector>
    {
        if a->n() !== b->n() {
            throw new InvalidArgumentException("Vector A requires "
                . (string) a->n() . " elements but Vector B has "
                . (string) b->n() . ".");
        }

        return static::quick(array_map("max", a->asArray(), b->asArray()));
    }
```

Removing the calls to `a->n(`), `b->n(`),` a->asArray(`), and `b->asArray()` fixes the problem. There is no issue when compiled with `internal-call-transformation: false`. I am inferring that there is something wrong with how Zephir handles instance methods in the static context, but I could be wrong.

Here is a link to the source file https://github.com/RubixML/Tensor/blob/master/tensor/vector.zep

And the test that fails/segfaults https://github.com/RubixML/Tensor/blob/master/tests/VectorTest.php

It may be important to note that `a->n(`), `b->n(`),` a->asArray(`), and `b->asArray()` access properties on the object using `this` which was the subject of a previous related bug https://github.com/phalcon/zephir/issues/1956

dreamsxin:
Hello!

* Type: bug fix
* Link to issue: #2000 

- [ ] I updated the CHANGELOG

Small description of change:

Thanks

### DIFF
--- a/ext/kernel/fcall.h
+++ b/ext/kernel/fcall.h
@@ -60,7 +60,7 @@ typedef enum _zephir_call_type {
 
 /* Saves the if pointer, and called/calling scope */
 #define ZEPHIR_BACKUP_THIS_PTR() \
-	zend_object *old_this_ptr = Z_OBJ(EG(current_execute_data)->This) ? Z_OBJ(EG(current_execute_data)->This) : NULL;
+	zend_object *old_this_ptr = Z_TYPE(EG(current_execute_data)->This) == IS_OBJECT ? Z_OBJ(EG(current_execute_data)->This) : NULL;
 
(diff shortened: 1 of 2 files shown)


sergeyklay:
Thank you!

PR #2039, opened 2020-01-02
Fix | Static Analyzers

AlexNDRmac:
Hello!

* Type: bug fix | code quality
* Link to issue:

- [X] I have checked that another pull request for this purpose does not exist
- [ ] I updated the CHANGELOG

Small description of change:

- Improved github actions to run all checks even if some check failed

Thanks

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -14,6 +14,9 @@ on:
 jobs:
   analysis:
 
+    strategy:
+      fail-fast: false
+
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -43,11 +43,13 @@ jobs:
       - name: Run PHP_CodeSniffer
         run: |
           ./phpcs --runtime-set ignore_warnings_on_exit true
+        continue-on-error: true
 
       - name: Run PHP CS Fixer
-        if: always()
-        run: ./php-cs-fixer fix --diff --dry-run -v
+        run: |
+          ./php-cs-fixer fix --diff --dry-run -v
+        continue-on-error: true
 
       - name: Run Shell Check
-        if: always()
         run: shellcheck .ci/*.sh
+        continue-on-error: true
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -43,13 +43,12 @@ jobs:
       - name: Run PHP_CodeSniffer
         run: |
           ./phpcs --runtime-set ignore_warnings_on_exit true
-        continue-on-error: true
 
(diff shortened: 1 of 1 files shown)


.github/workflows/analysis.yml
```diff
@@ -41,10 +41,13 @@ jobs:
           ./phpcs --version
 
       - name: Run PHP_CodeSniffer
-        run: ./phpcs
+        run: |
+          ./phpcs --runtime-set ignore_warnings_on_exit true
 
       - name: Run PHP CS Fixer
+        if: always()
```
ruudboon, line 34: Why not use:
```
continue-on-error: true
```
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

AlexNDRmac: Yes... it looks better. I'll fix this conditional, thanks :)

AlexNDRmac: But... job status always successful. Conditionals in steps doesn't change step status.
Main idea - is run all checks and see all reports from PHP tools... then developer can fix those problem with one fix commit.
So... I'll revert to explicit conditions :)

ruudboon: Ah, wasn’t expecting that status will change

.github/workflows/analysis.yml
```diff
@@ -41,10 +41,14 @@ jobs:
...
       - name: Run PHP_CodeSniffer
-        run: ./phpcs
+        run: |
+          ./phpcs --runtime-set ignore_warnings_on_exit true
 
       - name: Run PHP CS Fixer
-        run: ./php-cs-fixer fix --diff --dry-run -v
+        if: always()
+        run: |
+          ./php-cs-fixer fix --diff --dry-run -v
```
sergeyklay:
```diff
-        run: |
+        run: ./php-cs-fixer fix --diff --dry-run -v
-          ./php-cs-fixer fix --diff --dry-run -v
```

.github/workflows/analysis.yml
```diff
@@ -41,10 +41,14 @@ jobs:
           ./phpcs --version
 
       - name: Run PHP_CodeSniffer
-        run: ./phpcs
+        run: |
+          ./phpcs --runtime-set ignore_warnings_on_exit true
```
sergeyklay:
```diff
-        run: |
+        run: ./phpcs --runtime-set ignore_warnings_on_exit true
-          ./phpcs --runtime-set ignore_warnings_on_exit true
```

sergeyklay (changes requested):

sergeyklay:
Thank you

PR #2053, opened 2020-01-27
Fix missed PSR extension for CI workflow

AlexNDRmac:
Hello!

* Type: bug fix
* Link to issue:

In raising this pull request, I confirm the following:

- [X] I have checked that another pull request for this purpose does not exist
- [ ] I wrote some tests for this PR
- [ ] I updated the CHANGELOG

Small description of change:

- fixed missed PSR ext error for Linux CI builds with PHP 7.1-7.2

Thanks

PR #2054, opened 2020-01-27
Test and update workflow if ext-psr is loaded 

shivammathur:
This PR is to test and update workflow if `ext-psr` is loaded after `setup-php` step in the `build-linux` workflow . See https://github.com/phalcon/zephir/pull/2053#discussion_r371243844

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -8,6 +8,7 @@ on:
   pull_request:
     branches:
       - master
+      - development
 
 jobs:
   linux:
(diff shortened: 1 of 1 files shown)


sergeyklay:
@shivammathur https://github.com/phalcon/zephir/runs/410853128#step:3:41

PHP 7.1 - 7.2 without PSR extension. For example see https://github.com/phalcon/zephir/runs/410853128#step:3:41

shivammathur:
Yes, Fixing this.

shivammathur:
@sergeyklay You can merge this now. Fixed in `1.8.1`

sergeyklay (approved):

sergeyklay:
Thank you

PR #2085, opened 2020-04-27
fix duplicate definition with GCC 10

remicollet:
See https://github.com/ice/framework/pull/266

### DIFF
--- a/templates/ZendEngine3/php_project.h
+++ b/templates/ZendEngine3/php_project.h
@@ -50,7 +50,7 @@ ZEND_EXTERN_MODULE_GLOBALS(%PROJECT_LOWER%)
 #endif
 
 #ifdef ZTS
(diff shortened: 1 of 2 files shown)


sergeyklay:
@remicollet There are failed Linux tests due to changed template. Do not consider macOS ones (they are almost all broken and I turn them off today).

P.S. We use black box tests (in addition to Unit ones) to prove Zephir generates expected C-code and handles console commands. For more see https://github.com/phalcon/zephir/blob/development/unit-tests/sharness/README.md

remicollet:
Sorry, but I have no idea what you are talking about.

Please see the linked issue.
And provide a fix.

sergeyklay:
Fixed  in https://github.com/phalcon/zephir/pull/2087. Thank you for the patch.

PR #2087, opened 2020-04-27
Fix duplicate definition with GCC 10

sergeyklay:
Fixes: https://github.com/ice/framework/pull/266
Closes: https://github.com/phalcon/zephir/pull/2085

### DIFF
--- a/ext/test.c
+++ b/ext/test.c
@@ -216,6 +216,10 @@ zend_class_entry *test_unsettest_ce;
 zend_class_entry *test_usetest_ce;
 zend_class_entry *test_vars_ce;
 
+#ifdef ZTS
+ZEND_TSRMLS_CACHE_DEFINE()
+#endif
+
 ZEND_DECLARE_MODULE_GLOBALS(test)
 
 PHP_INI_BEGIN()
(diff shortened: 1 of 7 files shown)


ext/test.c
```diff
@@ -216,6 +216,10 @@ zend_class_entry *test_unsettest_ce;
 zend_class_entry *test_usetest_ce;
 zend_class_entry *test_vars_ce;
 
+#ifdef ZTS
+ZEND_TSRMLS_CACHE_DEFINE()
+#endif
```
sergeyklay, line 221: @remicollet 
Shouldn't it be changed as follows?

```c
/* implement standard "stub" routine to introduce ourselves to Zend */
#ifdef COMPILE_DL_%PROJECT_UPPER%
# ifdef ZTS
ZEND_TSRMLS_CACHE_DEFINE()
# endif
ZEND_GET_MODULE(%PROJECT_LOWER_SAFE%)
#endif
``` 

and bellow

```diff
static PHP_GINIT_FUNCTION(test)
{
- #ifdef ZTS
+ #if defined(COMPILE_DL_%PROJECT_UPPER%) && defined(ZTS)
 	ZEND_TSRMLS_CACHE_UPDATE();
 #endif
```

remicollet: indeed, even if don't see in which case `COMPILE_DL_%PROJECT_UPPER%` won't be define (excepted stupid edge case of people merge some ext in php-src)

sergeyklay: Thank you

PR #2114, opened 2020-08-24
Stubs for array

AlexNDRmac:
Hello!

When generating stubs with array declaration in params as `Foo[]` - stubs are duplicates.

```php
    /**
     * @param Foo[] $param
     */
    public function bar(array param)
    {
        let this->foos = param;
    }

but it generates this:
    /**
     * @param Foo[] $param
     * @param array $param
     * @param Foo  [] $param
     */
    public function bar(array $param)
    {
    }
```

* Type: bug fix

- [X] I have checked that another pull request for this purpose does not exist
- [X] I wrote some tests for this PR
- [X] I updated the CHANGELOG

Small description of change:

- Added support for generating stubs with square array declaration in docblock params

Thanks

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -28,11 +28,15 @@ jobs:
           tools: php-cs-fixer, phpcs
 
       - name: Run PHP_CodeSniffer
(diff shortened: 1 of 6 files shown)


ruudboon (approved):
LGTM

Issue #1977, opened 2019-10-17
Zephir stubs shouldn't add 'void' return type hint


sergeyklay:
As a far as Zephir supports PHP 7.0 the `stubs` command should generate methods with `void` return type hint:

```diff
- public function foo(): void
+ public function foo()
  {
  }
```

Currently this work so only if PHP 7.0 is used by generating stubs.

PR #2124, opened 2020-11-25
stub supports void type return value

fagai:
Hello!

* Type: bug fix | new feature
* Link to issue: https://github.com/phalcon/ide-stubs/pull/50
https://github.com/phalcon/zephir/issues/1977

- [ ] I updated the CHANGELOG

Small description of change:
I added it because I didn't consider the case where the return value is void type.

Thanks

### DIFF
--- a/Library/Stubs/Generator.php
+++ b/Library/Stubs/Generator.php
@@ -335,11 +335,6 @@ class Generator
                 $return = 'array';
                 ++$supported;
             }
-            
-            if (version_compare(PHP_VERSION, '7.1.0', '>=') && $method->areReturnTypesVoidCompatible()) {
-                $reutrn = 'void';
-                ++$supported;
-            }
 
             if (!empty($return) && $method->areReturnTypesNullCompatible()) {
                 if (version_compare(PHP_VERSION, '7.1.0', '>=')) {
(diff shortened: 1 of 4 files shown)


Library/Stubs/Generator.php
```diff
@@ -348,7 +348,10 @@ protected function buildMethod(ClassMethod $method, bool $isInterface, string $i
             if ($supported > 1) {
                 $return = '';
             }
+        } elseif (version_compare(PHP_VERSION, '7.1.0', '>=') && $method->isVoid()) {
```
AlexNDRmac, line 351: Better to use comparison with constant, its faster:

```php
elseif (PHP_VERSION_ID >= 70100 && $method->isVoid())
```

AlexNDRmac (changes requested):

Library/Stubs/Generator.php
fagai: @AlexNDRmac 
thank you.
I adjusted it for other parts, but is this better?

fagai:
https://github.com/phalcon/zephir/blob/development/Library/Stubs/Generator.php#L304-L309
Should I also remove this?

Library/Stubs/Generator.php
AlexNDRmac: Oh... I see :) This should be refactored... I think we can leave it as is for now. Thanks :)

AlexNDRmac:
> https://github.com/phalcon/zephir/blob/development/Library/Stubs/Generator.php#L304-L309
> Should I also remove this?

No, It should be removed with small refactoring later. Thank you.

fagai:
okay. Thank you.

AlexNDRmac (approved):

AlexNDRmac:
Thank you

PR #2135, opened 2021-01-25
Fix uninitialized named parameters field into the zend_fcall_info structure (PHP8 only)

ufoproger:
Hello!

* Type: bug fix
* Link to issue: (none)

- [ ] I updated the CHANGELOG

Small description of change:
PHP 8 allow using named arguments in functions call, so new field should be manually initialized. PHP RFC: [Named Arguments](https://wiki.php.net/rfc/named_params).


Jeckerson:
Nice! Seems to resolve seg fault.

Issue #2111, opened 2020-08-14
Support for php8


mruz:
Add php 8.0 support. We have feature freeze and beta1 now https://wiki.php.net/todo/php80.

It's disabled in composer/build. Can we enable (with allowed failures) and see what's failing?

Would be good if it's ready on Nov 26 2020 (GA).

Jurigag:
Yea i guess we will need to prepare stuff for it. I guess feel free anyone to figure out what needs to be done and use open collective funds for it.

@dreamsxin would have any time to at least do any review how lot of work this would require?

dreamsxin:
Changes:

Arg info must be defined.
Some zend api parameter type changed:  `zval` -> `zend_ object`.

Optimistic estimation: 2 weeks.

Jeckerson:
![Screenshot_20210125_130442](https://user-images.githubusercontent.com/3289702/105714648-84fba100-5f14-11eb-94ad-e8d40a59d1f2.png)

Work in progress.

mruz:
I'm receiving SIGABRT on kernel/fcall https://travis-ci.org/github/ice/framework/jobs/756221697
```zep

class Di extends Arr
{
    protected static di;

    public function __construct(array data = [])
    {
        parent::__construct(data);

        let self::di = this;
    }
```

```zep
class Arr implements \ArrayAccess, \Countable, \IteratorAggregate
{
    protected data = [] { get };

    public function __construct(array data = [])
    {
        let this->data = data;
    }
```

```
Program received signal SIGABRT, Aborted.
0x00007ffff26a8428 in raise () from /lib/x86_64-linux-gnu/libc.so.6
Thread 1 (Thread 0x7ffff7fc7ac0 (LWP 28558)):
...
#5  0x000000000085521d in zval_get_string_func (op=<optimized out>) at /tmp/php-build/source/8.0.1/Zend/zend_operators.c:932
#6  0x0000000000863f0e in zend_get_callable_name_ex (callable=<optimized out>, callable@entry=0x7fffffff98b8, object=<optimized out>) at /tmp/php-build/source/8.0.1/Zend/zend_API.c:3299
#7  0x000000000084f704 in zend_call_function (fci=fci@entry=0x7fffffff98b0, fci_cache=fci_cache@entry=0x7fffffff9890) at /tmp/php-build/source/8.0.1/Zend/zend_execute_API.c:717
[...]

PR #2148, opened 2021-02-17
#2111 - Add test case to simulate interface argument type

Jeckerson:
Hello!

* Type: bug fix | new feature | code quality | documentation
* Link to issue: #2111

```
<?php

class Test
{
    public function get(\Throwable $ex): void
    {
    }
}

$class = new Test();
try {
    $class->get(true);
} catch (\TypeError $ex) {
    echo $ex->getMessage();
}
```

Should throw exception, which doesn't happens at that moment.

Thanks

### DIFF
--- a/Library/ClassMethod.php
+++ b/Library/ClassMethod.php
@@ -1880,13 +1880,13 @@ class ClassMethod
          */
         $initCode = '';
         $code = '';
+        $requiredParams = [];
+        $optionalParams = [];
         if (\is_object($parameters)) {
             /**
              * Round 2. Fetch the parameters in the method.
              */
             $params = [];
-            $requiredParams = [];
-            $optionalParams = [];
             $numberRequiredParams = 0;
             $numberOptionalParams = 0;
             foreach ($parameters->getParameters() as $parameter) {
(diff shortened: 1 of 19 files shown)


tests/Extension/Interfaces/InterfaceMethodSignatureTest.php
```diff
@@ -0,0 +1,48 @@
...
+    public function testImplementInterfaceInMethodSignature(): void
+    {
+        $class = new ImplementInt();
+
+        $this->assertNull($class->get());
+        $class->set(1);
+        $this->assertSame(1, $class->get());
+
+        $error = '';
+        try {
```
AlexNDRmac: Better to use native PHPUnit methods for Exceptions instead of try/catch and string matches.

For example:
```php
$this->expectException(\TypeError::class);
$this->expectExceptionMessageMatches('/^must be of type int, bool given/');

$class->set(true);
```

tests/Extension/Interfaces/InterfaceMethodSignatureTest.php
```diff
@@ -0,0 +1,48 @@
...
+        } catch (\TypeError $exception) {
+            $error = $exception->getMessage();
+        }
+
+        $this->assertMatchesRegularExpression('/^must be of type int, bool given/', $error);
+
+        $this->assertSame(1, (new ImplementInterface())->get($class));
+
+        $error = '';
+        try {
```
AlexNDRmac: Same as above...

AlexNDRmac (changes requested):

Library/ClassMethod.php
```diff
@@ -2441,4 +2475,235 @@ public function areReturnTypesCompatible()
...
+            $className = $aliasManager->getAlias($className);
+        }
+
+        /**
+         * PSR
+         */
+        if (strpos($className, 'Psr') === 0 || strpos($className, '\Psr') === 0) {
+            $className = ltrim($className, '\\');
+
+            $psrHeaderFiles = [
```
AlexNDRmac, line 2607: My proposition - move this psr definition to separate file and use it as constant

AlexNDRmac (approved):

Library/ClassMethod.php
Jeckerson: Yes. This is just for fast tests, will be moved soon.

PR #2418, opened 2023-10-26
fast_add_function removed in 8.3

remicollet:
From UPGRADING.INTERNALS 
```
* The fast_add_function() has been removed, use add_function() that will
  call the static inline add_function_fast() instead.
```

This simple change allow to build Phalcon 5.4 against 8.3.0RC5 (but not tested)


remicollet:
Already in PR #2408

Jeckerson (approved):

PR #2444, opened 2025-01-17
Refactor CUFA implementation

Girgias:
Hello!

* Type:  code quality
* Link to issue:

- [x] ~I wrote some tests for this PR~ N/A
- [ ] I updated the CHANGELOG

Small description of change:

Use the named_params field to handle positional and named arguments for CUFA.

This saves on an unnecessary copy of the HashTable.

Long term project is to get rid of the `zend_fcall_info_args*()` APIs in Zend as they are not really intuitive, moreover there is no need to perform a copy of the arguments.

One question I have is if I need to do something to regenerate `ext/kernel` or will this be done at another time?

### DIFF
--- a/kernel/fcall.c
+++ b/kernel/fcall.c
@@ -610,15 +610,18 @@ int zephir_call_user_func_array_noex(zval *return_value, zval *handler, zval *pa
 
 	fci.size = sizeof(fci);
 	fci.object = fci_cache.object;
-	ZVAL_COPY_VALUE(&fci.function_name, handler);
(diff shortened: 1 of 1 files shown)


niden:
> One question I have is if I need to do something to regenerate ext/kernel or will this be done at another time?

@Girgias we can ignore it for now. It will be removed in the very near future.

Jeckerson (approved):

niden:
Thank you @Girgias

PR #2450, opened 2025-05-06
Converting To Use Local Variable Instead of Static Variable

Fenikkusu:
Hello!

* Type: bug fix
* Link to issue: N/A

- [x] I updated the CHANGELOG

Small description of change:

Taking a Stab at fixing the issue where calling a closure with a `use` statement within a loop always uses the last value:

IE: (Psudo Code)
```php
class Something extends Injectable {
    function doSomething(string! name, int value) {
        this->getDi() ->set(name, function() use (value) {
            return value;
        });
    }
}

var something = new Something();
something->doSomething("a", 1);
something->doSomething("b", 2);

// Current
di->get("a"); // Returns 2
di->get("b"); // Returns 2

// Should Be
di->get("a"); // Returns 1
di->get("b"); // Returns 2
```

I'm posting this here as I have no clue what I'm doing, and someone smarter than I may be able to make quick changes to this to finish it up. My latest change was to Method and stubs stopped building as a result and I don't have the time to keep digging ATM. Otherwise, I will keep working on it as time permits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format based on [Keep a Changelog](https://keepachangelog.com)
 and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
-
+- Fixed bug with Closures using the `use` Statement in a loop always returning the last loop.
 ## [0.19.0] - 2025-05-13
 ### Added
 - Added support of PHP `8.4` [#2440](https://github.com/zephir-lang/zephir/issues/2440), [#2443](https://github.com/zephir-lang/zephir/pull/2443)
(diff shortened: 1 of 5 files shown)


Fenikkusu:
Okay, I believe I have gotten this completed correctly now. I've split the original changes into multiple PRs specific to the needs of each ( #2456 & #2455 ), though one is no longer needed. Unfortunately, the Windows Errors are outside the scope of my focus at the moment, so someone else will need to handle those issues.

Fenikkusu:
Documenting: Attempting to compile Phalcon using this version appears to fail on my end. Once I confirm the issue, I will post an update.

Fenikkusu:
Never got this fixed and I forgot about it, closing.

PR #2455, opened 2025-06-03
Adding Block Manager To Allow Modifications of Headers In Expressions

Fenikkusu:
Hello!

* Type: new feature
* Link to issue: N/A

- [x] I updated the CHANGELOG

Small description of change:

As part of the changes I was "attempting" with #2450, at one point, I added these changes. This was originally done so I could have something like the below inside Closures.php. Eventually, my changes went another path, and these additions were no longer required. However, I still see them as having some value, so I wanted to go ahead and create a separate PR for these changes.

```php
$compilationContext->blockManager->add(
                <<<EOF
typedef struct _zend_closure {
	zend_object       std;
	zend_function     func;
	zval              this_ptr;
	zend_class_entry *called_scope;
	zif_handler       orig_internal_handler;
} zend_closure;
EOF
            );
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format based on [Keep a Changelog](https://keepachangelog.com)
 and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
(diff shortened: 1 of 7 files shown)


Fenikkusu:
This was more a sample (I think), so closing.

PR #2499, opened 2026-04-24
[BUG] Unable To Find External Dependencies on Case-Sensitive Systems

Fenikkusu:
Hello!

- [x] I updated the CHANGELOG

Many of my Zephir Projects extend or leverage Phalcon. I recently made a change to a project where I extended Phalcon\Support\Collection. I found, though, that if I called the "get" method, I was getting the error `called with a wrong number of parameters, the method has: 2, passed: 1` 

I loaded up Claude and started digging into the issue, and finally determined that this was the result of a combination of two bugs. The first bug (that this PR attempts to fix) is that when Zephir tries to load a class from `external-dependencies` in the config.json, it calls strtolower on the path. The issue is that on case-sensitive file systems, this check fails for Phalcon files because the raw files are capitalized. IE: The class `Phalcon\Support\Collection` was "mapping" to `/opt/cphalcon/phalcon/support/collection.zep`, which doesn't exist because the actual path was `/opt/cphalcon/phalcon/Support/Collection.zep`.

The second bug, which I'm not fixing at this time, but may still be required, is that it falls back to Reflection when it cannot find the class. In this case, Zephir uses the `isset` method in `Method::getNumberOfRequiredParameters`. The issue is that if the default value is null, as is the case in `Collection::get`, `isset` returns false.

This PR fixes the first issue, which should make the second issue a moot point. I have confirmed that after applying this fix, my project started to compile as expected. It is done in such a way as to allow backwards compatibility.


Fenikkusu:
[ZEPHIR_LOAD_EXTERNAL.md](https://github.com/user-attachments/files/27039306/ZEPHIR_LOAD_EXTERNAL.md)

Context From Claude Code

Fenikkusu:
I'm tired, and it's far past my bedtime. I will get the PR cleaned up tomorrow. I couldn't find (quickly) where the automated test for Compile is, so if someone wants to point me in the right direction, I'll get that handled too (I feel like I've asked that before...)

Jeckerson:
Closing in favour of https://github.com/zephir-lang/zephir/pull/2556

PR #2724, opened 2026-09-18
Implement `Buffer` kernel class for C array (#2721)

### LINKED ISSUE #2721: Contiguous Array or Class Properties backed as C Datatypes
### Feature

With [Tensor](https://github.com/RubixML/Tensor), the biggest performance challenge we have is dealing with the overhead of constantly converting PHP arrays to C arrays and back to PHP arrays. BLAS libraries require a contiguous buffer of data and we don't even bother to vectorize linear/elementwise operations (addition, subtraction, etc.) because unpacking and repacking the array dominates the runtime.

If we could somehow store the data in a C array, and then only materialize the PHP version when necessary, that would be a huge unlock.

It would enable multiple operations to be strung together without going through the PHP array intermediary and it would mean that we could vectorize ALL the operations instead of just a small subset of them where the overhead does not dominate the computation.

This could take the form of a new CArray datastructure/object that is backed by a contig C array with an API that bridges the gap between C land and PHP land or a way to store C datatypes on objects and then the developer could write code that bridges the gap. Or it could be something totally different - I haven't thought too much about the form it could take.

If it's something I can do myself as a kernel class then that would work as well.

### Example

_No response_

Jeckerson:
Hello!

- [x] I updated the CHANGELOG

Zephir benchmarks:
| subject | before | after | x
| :---: | :---: | :---: | :---:
| buffer read loop | 0.004995 | 0.136795 ops/μs | x27
| buffer write loop | 0.004474 | 0.215893 ops/μs | x48

Tensor benchmarks, `Tensor\BufferVector` prototype vs `Tensor\Vector`, same `.so`, n=1,000,000:

| action | Vector | BufferVector | x
| --- | --- | --- | ---
| add | 32.9 ms | 3.9 ms | x8.3
| multiply | 35.6 ms | 3.7 ms | x9.7
| dot | 16.8 ms | 0.40 ms | x42
| chain `a.add(b).multiply(c)` | 69.5 ms | 8.3 ms | x8.4
| storage | 16 Mb | 8.0 Mb | x2.1


niden (approved):

andrewdalpino:
Wow!

